PR #28, opened 2014-12-29
Local dev server using Vagrant, Ansible, and Nginx

mbland:
Folks I think might be interested, who may know enough about some of this stuff to set me straight, and who may merge on my behalf: @afeld @konklone @dlapiduz @NoahKunin

This configuration enables a developer to keep a server running both the internal and public versions of the Hub locally, without needing to switch between `jekyll serve` instances or running them on different ports. It doesn't match our prod environment exactly...yet...but it's close enough for development work, given that it just serves the static HTML.

The "internal mode" skirts the need to run a `google_auth_proxy` instance in development, but I've kept the `google_auth_proxy.yml` playbook as a record of how it might be used to deploy the it one day. (Current hangup: a better way of passing the `GOOGLE_AUTH_PROXY_CLIENT_ID` and `GOOGLE_AUTH_PROXY_CLIENT_SECRET` variables, especially since the Hub still uses values from my personal account.)

It's my first time playing with Vagrant and Ansible. Definitely open to suggestions. Long-term, it'd be nice to switch our prod deployment to Ansible, if that seems reasonable to folks. Also, this article definitely took a lot of the mystery out of getting this initial environment up and running: http://adamcod.es/2014/09/23/vagrant-ansible-quickstart-tutorial.html

As an aside: Longer-term, I'd like to create a `go` script to abstract a lot of the current setup and other commands, as per:
- http://www.thoughtworks.com/insights/blog/praise-go-script-part-i
- http://www.thoughtworks.com/insights/blog/praise-go-script-part-ii


mbland:
Just discussing Hub stuff with @gboone in general, and the topic of Vagrant and Ansible came up. At a high-level, we're thinking this configuration (for the Hub) would be a good working proof-of-concept to use as a reference for building a 18f.gsa.gov + Dashboard + Hub development environment.

Unless there are serious objections to this experiment, would anyone mind if @gboone pulled the trigger for me and merged this as-is?

PR #29, opened 2015-01-05
Extract Snippets module from JoinerImpl

mbland:
The goal is to extract a new gem to encapsulate the snippet-manipulation code. Eventually, I'd like to extract as many general-purpose gems as possible so that relatively little code remains in `_plugins`. The ideal is to have the remaining code serve as documentation of how to chain together Hub gems to generate one's own custom Hub instance.

@afeld Let me know if you'd like me to split these commits into separate PRs, if having them all-in-one is too much.


afeld:
You've been busy! :clap:

PR #39, opened 2015-01-08
Add "Edit data", "Hub Ops" to nav bar

mbland:
Also, the links from the footer have been removed, since they're now all in "Hub Ops".

@leahbannon @meiqimichelle @afeld: Suggestions welcome, of course!

Here's what the nav bar looks like for the internal version:
internal

And for the public version:
public

### DIFF
--- a/_data/nav_links.yml
+++ b/_data/nav_links.yml
@@ -1,8 +1,19 @@
 - name: Public Sites
   links:
-    - text: 18f.gsa.gov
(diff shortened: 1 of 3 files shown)


afeld:
The sidebar is getting pretty intense... but that's a problem for another day :football: :runner:

...a.k.a. https://github.com/18F/hub/issues/41

PR #52, opened 2015-01-11
Use ::TeamHub::Page from the new team_hub gem

mbland:
This is the first piece of Hub-specific infrastructure to be extracted into the [team_hub gem](https://github.com/18F/team_hub). More to come!

cc: @afeld @gboone @adelevie


adelevie:
LGTM.

+1 for good gem name
+1 for wiping out all that stuff thanks to gem extraction

PR #60, opened 2015-01-15
Fix /clips page

mbland:
@gbinal Confirmed the `/clips` page appears with these changes. After this is merged, I'll send you a PR to deploy to the public Hub.


gbinal:
Rock!  Thanks for catching my errors.

PR #61, opened 2015-01-15
Move deployment commands into ./go script

mbland:
The command strings in `deploy/fabfile.py` were getting a bit unwieldy; this is an attempt to abstract them in a more coherent way. Also, any tweaks to the deployment commands will not require any `fab [stop|start]` invocations.

Another approach might be to add separate scripts in `deploy/` for each individual `hookshot` instance, rather than sticking more commands in `./go`. Open to doing it that way as well, if folks prefer

BTW, I haven't relaunched the `hookshot` jobs to use this yet; I'll probably get around to that tomorrow.

cc: @gboone @konklone @afeld

### DIFF
--- a/go
+++ b/go
@@ -88,6 +88,34 @@ def ci_build
   build
 end
 
+AUTOMATED_DEPLOY_PULL_CMD = "git pull && git submodule update --remote"
+INTERNAL_BUNDLE_CMD = "/usr/local/rbenv/shims/bundle"
+PUBLIC_BUNDLE_CMD = "/opt/install/rbenv/shims/bundle"
+JEKYLL_BUILD_CMD = "exec jekyll b"
+JEKYLL_PUBLIC_CONFIG = "--config _config.yml,_config_public.yml"
+
+def deploy_submodules
+  exec_cmd(AUTOMATED_DEPLOY_PULL_CMD + ' && cd _data && ' +
+    '/usr/local/rbenv/shims/ruby ./import-public.rb && cd .. && ' +
+    'git add _data/private _data/public/ pages/private && ' +
+    'git commit -m \'Private submodule update\' && git push')
+end
+
+def deploy_internal
+  exec_cmd("%s && %s && %s %s && %s %s %s" % [
+    AUTOMATED_DEPLOY_PULL_CMD,
+    INTERNAL_BUNDLE_CMD,
+    INTERNAL_BUNDLE_CMD, JEKYLL_BUILD_CMD,
+    INTERNAL_BUNDLE_CMD, JEKYLL_BUILD_CMD, JEKYLL_PUBLIC_CONFIG])
+end
+
+def deploy_public
+  exec_cmd("%s && %s && %s %s %s" % [
+    AUTOMATED_DEPLOY_PULL_CMD,
+    PUBLIC_BUNDLE_CMD,
+    PUBLIC_BUNDLE_CMD, JEKYLL_BUILD_CMD, JEKYLL_PUBLIC_CONFIG])
+end
+
 COMMANDS = {
   :init => 'Set up the Hub dev environment',
   :update_gems => 'Execute Bundler to update gem set',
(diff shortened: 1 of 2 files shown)


go
```diff
@@ -88,6 +88,34 @@ def ci_build
...
+INTERNAL_BUNDLE_CMD = "/usr/local/rbenv/shims/bundle"
+PUBLIC_BUNDLE_CMD = "/opt/install/rbenv/shims/bundle"
+JEKYLL_BUILD_CMD = "exec jekyll b"
+JEKYLL_PUBLIC_CONFIG = "--config _config.yml,_config_public.yml"
+
+def deploy_submodules
+  exec_cmd(AUTOMATED_DEPLOY_PULL_CMD + ' && cd _data && ' +
+    '/usr/local/rbenv/shims/ruby ./import-public.rb && cd .. && ' +
+    'git add _data/private _data/public/ pages/private && ' +
+    'git commit -m \'Private submodule update\' && git push')
```
afeld: I think splitting the strings by command makes it more readable:

```ruby
def deploy_submodules
  exec_cmd([
    AUTOMATED_DEPLOY_PULL_CMD,
    'cd _data',
    '/usr/local/rbenv/shims/ruby ./import-public.rb',
    'cd ..',
    'git add _data/private _data/public/ pages/private',
    'git commit -m "Private submodule update"',
    'git push'
  ].join(' && '))
end
```

go
```diff
@@ -88,6 +88,34 @@ def ci_build
...
+end
+
+def deploy_internal
+  exec_cmd("%s && %s && %s %s && %s %s %s" % [
+    AUTOMATED_DEPLOY_PULL_CMD,
+    INTERNAL_BUNDLE_CMD,
+    INTERNAL_BUNDLE_CMD, JEKYLL_BUILD_CMD,
+    INTERNAL_BUNDLE_CMD, JEKYLL_BUILD_CMD, JEKYLL_PUBLIC_CONFIG])
+end
```
afeld: I don't use string replacement very often... the string to be replaced is confusing to look at, though I like how you laid out the arguments to make things more obvious.

I took a crack at refactoring, but thinking I made things worse.  Anyway, here it is... take it or leave it.

```ruby
def join_args(args)
  args.join(' ')
end

def join_commands(commands)
  commands.join(' && ')
end

def bundle_cmd(command=nil)
  args = [INTERNAL_BUNDLE_CMD, command].compact
  join_args(args)
end

def deploy_internal
  command = join_commands([
    AUTOMATED_DEPLOY_PULL_CMD,
    bundle_cmd,
    bundle_cmd(JEKYLL_BUILD_CMD),
    bundle_cmd(join_args([JEKYLL_BUILD_CMD, JEKYLL_PUBLIC_CONFIG]))
  ])
  exec_cmd(command)
end
```

go
```diff
@@ -98,19 +126,32 @@ COMMANDS = {
...
 
-  padding = COMMANDS.keys.max_by {|i| i.size}.size + 2
-  COMMANDS.each do |name, desc|
-    puts "  %-#{padding}s#{desc}" % name
+  COMMAND_SECTIONS.each do |command_section|
+    puts "\n#{command_section[:section]}:"
+    commands = command_section[:commands]
+    padding = commands.keys.max_by {|i| i.size}.size + 2
+    commands.each {|name, desc| puts "  %-#{padding}s#{desc}" % name}
   end
```
afeld: This could use some explanation via inline comments... or a descriptive method.

go
```diff
@@ -120,7 +161,8 @@ command = ARGV.shift
 usage if ['-h', '--help'].include? command
 
 command = command.to_sym
-unless COMMANDS.member? command
+ALL_GO_COMMANDS = COMMAND_SECTIONS.map {|i| i[:commands].keys}.flatten
+unless ALL_GO_COMMANDS.member? command
```
afeld: Ditto.

gboone: Agree, I'm a little confused about all of what's going on in this last section.

mbland:
@afeld @gboone I've pushed a new commit incorporating y'all's feedback. Let me know if you've further suggestions on how I can make the flow more clear.

Also, forgot to mention in the commit message, but I've updated `init` and `update gems` slightly.

PR #64, opened 2015-01-15
Various tests, utils, and refactorings

mbland:
If this is too much for one PR, happy to split out orthogonal pieces. All of this is refactoring and testing with an eye towards moving more functionality into the [team_hub gem](https://github.com/18F/team_hub).

@afeld I know you're going on vacation, so no rush.
@adelevie In case you have cycles, love to see what you think of some of my tricks, particularly regarding `TEST_TMPDIR` and  `_test/temp_file_helper.rb`.
@gboone In case you'd like to take a look.


_data/import-public-snippets.rb
```diff
@@ -25,16 +25,19 @@
...
 
-site = ::Jekyll::Site.new ::Jekyll::Configuration::DEFAULTS
-site.read_data('.')
-joiner_impl = ::Hub::JoinerImpl.new site
+team_filename = File.join('private', 'team.yml')
+team = SafeYAML.load_file(team_filename, :safe=>true)
+unless team
+  puts "Failed to parse #{team_fileanme}"
+  exit 1
+end
```
afeld: Minor: maybe put all of this in a `get_team` method or something?

_test/private_assets_copy_team_images_test.rb
```diff
@@ -0,0 +1,72 @@
...
+# <https://creativecommons.org/publicdomain/zero/1.0/>.
+#
+# @author Mike Bland (michael.bland@gsa.gov)
+
+require_relative "test_helper"
+require_relative "../_plugins/private_assets"
+require_relative "site"
+require_relative "temp_file_helper"
+
+require "minitest/autorun"
```
afeld: If a bunch of these requires are common to many test files, might as well move them into the `test_helper`.

afeld:
"Vacation" is that thing where you work from somewhere other than you live, right? Or is that just me? :snowboarder: :snowflake: :wink: 

Just a couple minor notes, but otherwise :shipit:

_data/import-public-snippets.rb
mbland: Done; called it `load_team_data`.

_test/private_assets_copy_team_images_test.rb
mbland: I'm big on explicit, narrowly-scoped `require` statements in each file, to reduce dependencies and the chance that removing a dependency in one file will break another. Maybe that's paranoid on my part, but it helps me sleep at night! :-)

PR #67, opened 2015-01-18
Extract and use test_temp_file_helper gem

mbland:
Makes use of the new [test_temp_file_helper gem](https://github.com/18F/test_temp_file_helper).

cc: @afeld @gboone @adelevie


afeld:
Corresponds to https://github.com/18F/team_hub/pull/1.

Issue #42, opened 2015-01-08
Investigating client-side search


mbland:
Something along the lines of:

https://github.com/blog/1939-how-github-uses-github-to-document-github

afeld:
https://github.com/PascalW/jekyll_indextank may be useful.

@jglovier @benbalter @gjtorikian We're looking to replicate how you implemented search (and for a similar purpose) as what you described in the blog post above...any chance the code that generates (and ideally, consumes) https://help.github.com/search/search.json is open source? If not, any interest in collaborating on a gem?

gjtorikian:
It's all totally open-source! 

We use [lunrjs](https://github.com/olivernn/lunr.js) to build the search. A gem does exist to [easily support it in Jekyll](https://github.com/slashdotdash/jekyll-lunr-js-search), but, it hasn't been updated for Jekyll 2.0, and is largely abandoned, I think.

But that's totally fine. Without the gem dependency, we can do all of this in native Jekyll, which makes it much easier to deploy to GHPages. Here's what we do:

* Include lunrjs and [mustache.js](https://github.com/janl/mustache.js) as assets
* Create a blob of HTML that can be passed to lunrjs. This is an HTML template that's written with mustache:

   ``` html
<script id="quicksearch-results-template" type="text/mustache">
{% raw %}
  {{#entries}}
    <tr class="article">
      <td class="info">
        <a href="{{url}}" class="js-articles-quicksearch-link">{{title}}</a>
      </td>
    </tr>
  {{/entries}}
{% endraw %}
</script>
   ```
* Set lunrjs up (above the template definition):
   ``` javascript
    <script type="text/javascript">
    $(function() {
      $('#quicksearch-query').lunrSearch({
        indexUrl: '{{ site.baseurl }}/search/search.json',
        quickSearchResults:  '.autocomplete-results',
        quickSearchEntries:  '.result-group',
        quickSearchTemplate: '#quicksearch-results-template',

        searchResults: '.search-container',
        searchEntries: '.search-results',
        searchTemplate: '#search-results-template'
      });
    });
  </script>
   ``` 
* Define some sort of *search.json* file. Ours looks something like this:
  ``` json
---
---
{

  "entries": [
    {% for article in site.dotcom_articles %}
    {
      "title": "{{ article.title }}",
      "category": "{{ page_category }}",
      "category_url": "{{ page_category_url }}",
      "url": "{{ site.baseurl }}/articles/{{ article.title | slugify }}",
[...]

PR #69, opened 2015-01-21
create an API for pages

afeld:
This is one part of https://github.com/18F/hub/issues/42: generating a JSON file with data about the site's pages. This PR temporarily includes two approaches:

* A Liquid template to generate the JSON, as suggested by @gjtorikian in https://github.com/18F/hub/issues/42#issuecomment-69274205
    * Pro: doesn't require any plugin.
    * Con: the manual escaping of the values is fragile and messy. We have JSON encoding built into Ruby, so why not use it.
* An addition to our API plugin to generate `/api/pages/`
    * Pro: will be easier to extract to a gem
    * Con: it requires a plugin, which means it won't work on GitHub Pages

My plan is to go with the latter, but there are a couple outstanding issues:

* [x] The two approaches are currently generating slightly different content, though I'm having trouble figuring out why they differ.
* [x] The templates aren't being evaluated (in either case), so I'm getting JSON that looks like this:

    ```json
    {
      "title": "",
      "url": "/wg/",
      "body": "Working Groups  {% for wg in site.data.working_groups %} ..."
    }
    ```

Halp!


_plugins/api.rb
```diff
@@ -85,6 +93,38 @@ def self.generate_locations_endpoint(site)
...
+
+    def self.pages(site)
+      site.pages.select {|page| %w(.html .md).include?(page.ext) }
+    end
+
+    def self.get_output(site, page)
+      do_render(site, page)
+      filterer = Filterer.new
+      filterer.condense(filterer.strip_html(page.content))
+    end
```
afeld: @parkr @mattr- I know you don't have a lot of context here, but in short: how do I access the fully rendered content of a `Page` within a plugin? The code above is close, though the liquid tags are still present. Any ideas? I looked through some existing Jekyll plugins, but didn't see anything that attempted this.

Side note: even with prodding around the Jekyll source code, getting this far was difficult; the `Page`s having an implicit state of being `#render`ed or not is pretty confusing. Going to try and take what I learn(ed) here and do a PR on Jekyll to make the `content`/`output`/`render` methods idempotent. Just FYI :wink:

mbland: Why do you need to generate the output in the plugin itself? How will it be consumed? I was assuming that the normal Jekyll site generation would take place, then the indexing would happen afterwards.

parkr: @afeld Prior art for your last point: https://github.com/jekyll/jekyll/pull/2901

Generators are run *before* the render step. So generators can create Liquid code and pages and stuff that print out later with the fully-rendered stuff. That said, pages cannot reference each other :(

There is a `page.render()` you can use in a pinch, but it'll slow down the build because those pages will be rendered twice.

afeld: > I was assuming that the normal Jekyll site generation would take place, then the indexing would happen afterwards.

That could work. Ideally it wouldn't require any monkey-patching or extra build step, but that might be good-enough-for-now.

> There is a `page.render()` you can use in a pinch

Yep, using that a few lines above... is there a step or option or something I'm missing?

> it'll slow down the build because those pages will be rendered twice.

I'm ok with that for an MVP, but good to know!

parkr: You'd have to also render the pages you depend on. If at all possible, move to collections. If not, name your api project something that will be sorted last.

afeld: > If at all possible, move to collections.

What would that do?

parkr: Collections are rendered before pages.

afeld:
I've done it!

young frankenstein

This approach shoehorns in an `after_render` hook on the `site`, so that the JSON generator (note, not a `Jekyll::Generator`) can be run after all the other pages have been rendered.

afeld:
Damn you @benbalter, stop stealing all of my ideas years in advance! http://stackoverflow.com/a/15572209/358804

afeld:
Going with https://github.com/18F/hub/pull/80 instead.

afeld:
@mbland Updated to use the gem, to split out the backend changes from https://github.com/18F/hub/pull/80.

_plugins/api.rb
```diff
@@ -85,6 +85,11 @@ def self.generate_locations_endpoint(site)
         'Index of team members by location code', data)
     end
 
+    def self.generate_pages_endpoint(site)
+      # file created through jekyll_pages_api gem, so just need to return information for the index page
```
mbland: I know this sounds funny, but would you mind wrapping this at 80 chars?

afeld: Haha, I am anti hard line wrapping :wink:

_test/pages_api_test.rb
```diff
@@ -0,0 +1,59 @@
...
+require_relative 'test_helper'
+
+require 'minitest/autorun'
+
+module Hub
+  class PagesApiTest < ::Minitest::Test
+    # before all
+    # TODO isolate the tests better
+    BUILD_DIR = File.join(Dir.pwd, '_test', 'tmp')
+    `bundle exec jekyll build --destination #{BUILD_DIR}`
```
mbland: Hmm. This is an interesting way to run a large test... I'll merge it for now, but we should probably sync up over video to discuss it sometime.

mbland:
Just a couple minor comments; feel free to :shipit: after addressing, or ping it back to me.

_plugins/api.rb
mbland: Well, I'm strongly for it, and since it's the existing style in this project... ;-) I'll do it and merge.

_test/pages_api_test.rb
afeld: I added this file of tests before moving the plugin to a separate gem, so honestly wouldn't be opposed to removing this entire file. I am [doing the same thing in an integration test](https://github.com/18F/jekyll_pages_api/blob/6df7299eca91ad22de2fdad837b3e6efc954f50e/spec/integration_spec.rb#L27) in jekyll_pages_api, though there are now additional unit tests in there.

PR #70, opened 2015-01-21
Prefix ./go script with ruby path, not bundle path

mbland:
@afeld @gboone A quick fix for the `./go` script deployment setup. Confirmed that it works on the internal Hub, and the new task is running on 18f.gsa.gov. When I push a new batch of public snippets (or maybe even locations!), that'll confirm that the Public Hub deployment is still operational.

### DIFF
--- a/deploy/fabfile.py
+++ b/deploy/fabfile.py
@@ -33,24 +33,24 @@ import fabric.api
 #   fab [command] --set instance=public"
 INSTANCE = fabric.api.env.get('instance', 'internal')
 
-INTERNAL_BUNDLE_CMD = "/usr/local/rbenv/shims/bundle"
-PUBLIC_BUNDLE_CMD = "/opt/install/rbenv/shims/bundle"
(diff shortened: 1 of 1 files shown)


afeld:
I think what you want is `bundle exec ruby ./go`... the shims *should* already be in the `PATH`, and doesn't limit running the deployment to people who use rbenv (which I don't).

mbland:
Actually, the shims aren't in the path for the `hookshot` task, hence the full paths. Since this script is specific to our deployment, I'm inclined to just go with this and not dig deeper. But, I'm open to persuasion.

afeld:
I think you'll still want to leverage bundler to restrict the gem versions...maybe `/opt/install/rbenv/shims/bundle exec /opt/install/rbenv/shims/ruby ./go` then? ::shudders::

mbland:
The `bundler` paths are already baked into the script. :-)

afeld:
You and your ideas.

Issue #77, opened 2015-01-22
Revamp team index page


mbland:
The `/team` page is currently generated by `_plugins/team.rb`. Ideally it should be generated straight from a Markdown page in `/pages`.

mbland:
Hey @meiqimichelle, the team index page is now `pages/team.html`. You're free to work your magic on it now! :-)

PR #78, opened 2015-01-23
site.data['team'] kept as Array; new /team/ index

mbland:
Per #77. **DO NOT MERGE YET.**

Confirmed that existing tests pass and the output from `./go serve` looks as expected. However, I'd prefer #74 go in first, and I need to add some tests for `api.rb` and `cross_referencer.rb`. Just wanted to show progress on #77 was forthcoming.

cc: @meiqimichelle

### DIFF
--- a/_plugins/cross_referencer.rb
+++ b/_plugins/cross_referencer.rb
@@ -36,12 +36,14 @@ module Hub
 
     def initialize(site_data)
       @site_data = site_data
+      @team = {}
+      @site_data['team'].each {|i| @team[i['name']] = i}
     end
 
     # Cross-references geographic locations with team members.
     def xref_locations_and_team_members
       locations = {}
-      @site_data['team'].each do |unused_name, member|
+      @site_data['team'].each do |member|
         member_location = member['location']
         if member_location
           unless locations.member? member_location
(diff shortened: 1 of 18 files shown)


mbland:
OK, I think this may be ready to merge now.

I didn't cover every last line where I needed to change `site.data['team'].values` to `site.data['team']`, but I covered a lot and laid the ground work to both extract some parts of `cross_referencer.rb` into [team_hub](https://github.com/18F/team_hub) and to refactor other parts to eliminate some currently-slightly-hairy logic. Plus, I don't want @meiqimichelle's #77 to block on me any longer.

cc: @afeld @meiqimichelle @gboone

_plugins/cross_referencer.rb
```diff
@@ -28,6 +28,118 @@ def self.build_xrefs(site_data)
...
+    # Creates an index of +collection+ items based on +key+.
+    #
+    # @param collection [Array<Hash>] collection from which to build an index
+    # @param key [String] property used to build the index
+    # @return [Hash]
+    def self.create_index(collection, key)
+      index = {}
+      collection.each {|i| (index[i[key]] ||= Array.new) << i if i[key]}
+      index
+    end
```
afeld: Nice.

PR #80, opened 2015-01-26
add client-side search

afeld:
Closes https://github.com/18F/hub/issues/42.

This is a superset of https://github.com/18F/hub/pull/69, but the generator has been moved to [the jekyll_pages_api gem](https://github.com/18F/jekyll_pages_api), and there is now client-side code to do searches, written in Angular (which was a fairly arbitrary choice).

demo

Note that you can ignore anything under `assets/components/`, as those are all just vendored files.


afeld:
A couple other things to note:

* Keyboard shortcuts are supported.
    * `/` to jump to the search box
    * `UP`/`DOWN` arrows then `ENTER` to navigate to a page
* I spent a lot more time on the search interaction than the quality of the results
    * More pages are included than we probably want, e.g. the 404 page. We can make a blacklist, but not sure if there's a more clever way to filter them.
    * Posts aren't included – see https://github.com/18F/jekyll_pages_api/issues/2.
    * The `boost`s (scoring) could probably use tweaking.

afeld:
Ok, merged `master` and removed the unnecessary files – should be much easier to grok now!

afeld:
Also, I normally don't advocate for rebasing/squashing commits once they're in a PR, but would understand if that were preferable in this case to remove the (essentially) duplicate commits from the split out PRs... let me know.

mbland:
Yeah, go ahead and squash and rebase and `push -f` as makes sense.

assets/_sass/_search.scss
```diff
@@ -0,0 +1,36 @@
+// https://github.com/mauriciogentile/angular-livesearch/blob/1f4e357ab3b50701eeedcfb45c12cd901d612087/example/styles.css#L31-L65
+ul.searchresultspopup {
```
msecret: Why do you need the ul here? Why is .searchresultspopup not specific enough?

assets/_sass/_search.scss
```diff
@@ -0,0 +1,36 @@
...
+  border: #a4bed4 1px solid;
+  margin-top: 0px;
+  padding: 0px;
+  z-index: 99999!important;
```
msecret: Its pretty dangerous to start using z-indexes without organizing them. you could end up having 9999999 next time. I suggest making sass variables for z-indexes.

_includes/head.html
```diff
@@ -78,6 +78,10 @@
 ================================================== -->
 <script src="{{ site.baseurl }}/assets/js/jquery-2.1.1.min.js"></script>
 <script src="{{ site.baseurl }}/assets/js/responsiveslides.min.js"></script>
+<script src="{{ site.baseurl }}/assets/components/angular/angular.min.js"></script>
```
msecret: I think js scripts should be loaded at the bottom of the page, unless using the async attribute.

_layouts/bare.html
```diff
@@ -21,8 +21,25 @@
...
   </div>
 
-  <div class="bare-content" role="main" itemscope itemprop="mainContentOfPage">
-    {{ content }}{% include edit_link.html %}
+  <div ng-app="hubSearch" ng-controller="SearchController" class="bare-content" role="main" itemscope itemprop="mainContentOfPage">
+    {% raw %}
+      <div>
+        <live-search id="search1" type="text"
+          live-search-callback="searchCallback"
+          live-search-item-template="<a href='{{result.page.url}}'>{{result.page.title || result.page.url}}</a>"
```
msecret: That a tag could probably use a title tag for 508 compliance.

_layouts/bare.html
```diff
@@ -21,8 +21,25 @@
     {% include nav.html %}
   </div>
 
-  <div class="bare-content" role="main" itemscope itemprop="mainContentOfPage">
-    {{ content }}{% include edit_link.html %}
+  <div ng-app="hubSearch" ng-controller="SearchController" class="bare-content" role="main" itemscope itemprop="mainContentOfPage">
+    {% raw %}
+      <div>
+        <live-search id="search1" type="text"
```
msecret: The use of an id here concerns me. Why is there a "1"? What if another search for has to be included on the page? Could this be made modular instead?

assets/_sass/_search.scss
afeld: Honestly, I copied-and-pasted this file from the URL above – didn't really spend any time cleaning it up. Worth a pass to delete anything we don't need, I gueeeeesssss :expressionless:

assets/js/search.js
```diff
@@ -0,0 +1,101 @@
...
+// based on https://github.com/angular/angular.js/blob/54ddca537/docs/app/src/search.js#L198-L206
+ngHub.factory('searchUi', function($document) {
+  var giveSearchFocus = function() {
+    var input = angular.element('#search1')[0];
+    input.focus();
+  };
+
+  var onKeyDown = function(event) {
+    // forward slash
+    if (event.keyCode === 191 && document.activeElement.tagName !== 'INPUT') {
```
msecret: I might be totally wrong here, but I could imagine tagName returning results in different cases depending on browsers. I'd recommend a simple toLower here to eliminate any possible inconsistencies.

assets/js/search.js
```diff
@@ -0,0 +1,101 @@
+---
+---
```
mbland: Might want to document that the `---` lines are necessary to ensure that `{{ site.baseurl }}` is substituted by Jekyll.

And, in general, mind giving a brief overview of the flow of the callbacks here? Also, each of the functions could use a brief descriptive comment.

_layouts/bare.html
afeld: What should be in there? Seems like it's not necessary since the text is displayed:

http://stackoverflow.com/questions/5997371/is-the-title-attribute-for-links-required-for-508-compliance

/cc @nickbristow

assets/js/search.js
afeld: [Supposedly not](https://developer.mozilla.org/en-US/docs/Web/API/Element.tagName#Notes), but couldn't hurt. Good catch!

mbland:
Gonna merge now, but opening an issue to address @msecret's comments and the observation that the search box appears to only allow up to nine characters.

PR #81, opened 2015-01-26
Fix unordered list item styles across the site

mbland:
These styles used to only affect snippets. It only affects the first two
nesting levels; my CSS knowledge is not sufficient to fix lists nested more
deeply.

@gboone This'll make the new internal publishing guidelines look more normal.

### DIFF
--- a/assets/_sass/_hub.scss
+++ b/assets/_sass/_hub.scss
@@ -11,6 +11,22 @@
   }
 }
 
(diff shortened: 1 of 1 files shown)


afeld:
> my CSS knowledge is not sufficient to fix lists nested more deeply

I'd have to see what you're trying to do, but happy to screenshare to pair on this if you like.

PR #82, opened 2015-01-26
add Bower components for #80

afeld:
As a way to split up https://github.com/18F/hub/pull/80, this adds the client-side components to be used for the client-side search. You can see the changes without the actual assets in this diff: https://github.com/18F/hub/compare/e9811d277e252df829deda5eed3a633926bde375...66138ac351c65e3f2a8e56f5fdf2f2ae1a18dfd6. I also reduced the number of committed files by using [bower-installer](https://github.com/blittle/bower-installer) to only copy in the relevant scripts.

### DIFF
--- a/go
+++ b/go
@@ -58,6 +58,12 @@ def init
   exec_cmd 'bundle install'
 end
 
+def update_components
+  exec_cmd 'npm install -g bower bower-installer'
+  exec_cmd 'bower update'
+  exec_cmd 'bower-installer'
+end
+
 def update_gems
   exec_cmd 'bundle update'
   exec_cmd 'git add Gemfile.lock'
@@ -167,6 +173,7 @@ CommandGroup.add_group(
   'Development commands',
   {
     :init => 'Set up the Hub dev environment',
+    :update_components => 'Execute bower-installer to update Bower components. Requires Node.js.',
     :update_gems => 'Execute Bundler to update gem set',
     :test => 'Execute automated tests',
     :serve => 'Serves the internal hub at localhost:4000',
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,0 +1,3 @@
+{
+  "directory": "assets/components"
+}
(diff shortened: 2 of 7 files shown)


afeld:
This is very strange...it seems to be showing the commits out of order. Bug in GitHub?

go
```diff
@@ -58,6 +58,12 @@ def init
   exec_cmd 'bundle install'
 end
 
+def update_components
```
mbland: Should this be installed from `init`?

go
```diff
@@ -58,6 +58,12 @@ def init
...
+  exec_cmd 'npm install -g bower bower-installer'
```
mbland: Should this line go in `init`, perhaps?

go
```diff
@@ -167,6 +173,7 @@ CommandGroup.add_group(
   'Development commands',
   {
     :init => 'Set up the Hub dev environment',
+    :update_components => 'Execute bower-installer to update Bower components. Requires Node.js.',
```
mbland: If this is Bower-specfic, should it be called `update_bower`?

mbland:
Mostly looks good, but I have a few questions first. Also, will this mean that Node/Bower is required to develop/deploy the Hub now? If so, `README.md` needs to be updated.

go
afeld: I was debating it, but this keeps the barrier to entry low. Since the dependencies are being checked in, Node/Bower isn't actually needed unless installing new packages or updating existing ones.

go
afeld: Doesn't have to be Bower-specific (e.g. it could later include a minification step or something), but I don't feel strongly one way or another.

afeld:
> will this mean that Node/Bower is required to develop/deploy the Hub now?

See comment above – only for adding/updating Bower components at the moment, though there would be other advantages if we expected it to be installed. Not proposing that at the moment.

PR #87, opened 2015-02-01
cross_referencer tests and refactoring

mbland:
The `CrossReferencerImpl` is now completely covered by tests, and I made some improvements to boot. Once this goes in, I'll extract these bits into [team_hub](https://github.com/18F/team_hub).

cc: @afeld

### DIFF
--- a/_plugins/canonicalizer.rb
+++ b/_plugins/canonicalizer.rb
@@ -50,6 +50,10 @@ module Hub
             sort_by_last_name! wg[member_type] if wg.member? member_type
           end
         end
+
+        site_data['team'].each do |member|
+          (member['working_groups'] || []).sort_by! {|wg| wg['name']}
+        end
       end
     end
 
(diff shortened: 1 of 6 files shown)


afeld:
My overall feedback is that it would be nice to not need all of the `|| []` everywhere... maybe something like this in the initialization, so you only need to do it once?

```ruby
site_data['snippets'] ||= []
```

mbland:
I'll give that a go and ping again when ready.

mbland:
Hey @afeld, I haven't updated this yet, but I should've mentioned: I originally left things `nil` so that the Liquid templates could check (for example) `{% if member.working_groups %}` before deciding whether or not to render the "Working Groups" section of the page, versus `{% if member.working_groups.size != 0 %}` or rendering the "Working Groups" headline without any actual content.

Hence, a question: Keep things as-is, or add a filtering pass to remove empty Arrays and Hashes from all objects after Joining/Cross-referencing? (Might be something to add to `hash-joiner`, a recursive attribute pruner.)

afeld:
Gotcha. No strong feelings!

mbland:
Cool. Then do you mind merging as-is? Then I can yank this code into [`team_hub`](https://github.com/18F/team_hub) and revisit the decision there.

mbland:
Starting to address this in 18F/hash-joiner#7.

PR #94, opened 2015-02-05
Add team map to locations page

shawnbot:
This PR introduces a team map to the `/locations/` page, and some conventions for per-page scripts and CSS. Namely, the head and scripts includes now look for `page.stylesheets` and `page.scripts`, respectively, and add the relevant `<link>` and `<script>` tags for each one. You can see this in action on the [locations page](https://github.com/18F/hub/blob/feature/team-map/pages/locations.html#L4-L9):

```yaml
scripts:
- /assets/js/vendor/d3.v3.min.js
- /assets/js/vendor/queue.v1.min.js
- /assets/js/vendor/topojson.v1.min.js
- /assets/js/map.js
stylesheets:
- /assets/css/map.css
```

[map.js](https://github.com/18F/hub/blob/feature/team-map/assets/js/map.js) then looks for `#team-map`, loads some [US states in TopoJSON](https://github.com/18F/hub/blob/feature/team-map/assets/data/us-states.json) and the team data from `/api/team/`, groups members by location (these are [hard-coded](https://github.com/18F/hub/blob/feature/team-map/assets/js/map.js#L9-L20)), and makes a map.

Each location on the map is wrapped in an `<a>` that naïvely links to `/locations/{code}/`, some of which may not exist. We could eliminate the hard-coded locations by moving their coordinates and city names into a new YAML data file, and exposing that info in the `/api/locations/` endpoint. cc: @mbland


mbland:
Diggin' it! :-) Before I forget: @msecret, would you be able to take a peek at the JS stuff? My eyes aren't good enough. (If only it were C++...) :-P

Other bits:

As I mentioned in Slack, the anchors pointing to location endpoints can instead point to targets on the locations page, i.e. `locations#aus` instead of `locations/aus`. Then I can do a little bit of work and eliminate the `_plugins/locations.rb` file.

As I also mentioned in Slack, you can create `_data/locations.yml` (in a later PR if you want), and we have the technology via the `hash-joiner` Gem to merge that info with the info harvested from the `location:` properties of individual team members, and expose the new data through the `api/locations/` endpoint as well.

And, finally, as I mentioned you-know-where, we could lower-case the word 'Snippets' in the failing test to get it passing again.

mbland:
Hey @shawnbot, I tried to `git pull`, then `git -i rebase` this onto `origin/master`. Eventually got it to merge after removing a few redundant commits already in `master` and the "no more <br> tags" commit ( c3190d3), but now the map shows up as `Error: undefined`. Tried moving the scripts from `assets/js/vendor/` to `_assets/javascripts/js` as per one of Aidan's changes, but no dice.

I can try to dig in later, but feel free to beat me to it.

mbland:
So I figured out the problem: after merging `master` back into this branch, since @gboone's JSON content-type fix from #99, the `/api/team` endpoint now serves `/api/team/api.json` via an nginx config tweak. However, `jekyll serve` doesn't know to do this. Hence, when running `./go serve` and hitting `http://localhost:4000/locations/`, the map wasn't loading properly.

I pushed a fix to this branch/PR in b9a4387. The merge history looks ugly, and I don't know if the parallel commits'll get filtered out properly. I'll leave the decision of whether or not to `git rebase -i origin/master` and `git push -f` to @shawnbot.

mbland:
Oh, forgot one of the other benefits of a rebase and `push -f`: the scope of changed files in the PR is reduced, focusing the review.

_includes/team_members.html
```diff
@@ -4,10 +4,13 @@
       <a href="{{ site.baseurl }}/team/{{ member.name }}"><img class="team-member-thumb" src="{{ site.baseurl }}/{{ member.image }}" alt="Thumbnail photo of {{ member.full_name }}"></a>
     </div>
     <div class="team-member-info">
-      <a href="{{ site.baseurl }}/team/{{ member.name }}">{{ member.full_name }}</a>{% if member.location %} - <a href="{{ site.baseurl }}/locations/{{ member.location }}">{{ member.location }}</a><br/>{% endif %}
-      <p>{% if member.projects and member.projects.length != 0 %}
-      Projects: {% for p in member.projects %}<a href="{{ site.baseurl }}/projects/{{ p.name }}">{{ p.project }}</a>{% unless forloop.last %}, {% endunless %}{% endfor %}{% endif %}
+      <a href="{{ site.baseurl }}/team/{{ member.name }}">{{ member.full_name }}</a>{% if member.location %} -
```
msecret: title tag for accessibility

assets/js/map.js
```diff
@@ -0,0 +1,203 @@
+(function(exports) {
+
+  var map = d3.select("#team-map")
+        .classed("loading", true),
+      status = map.append("p")
+        .html("Loading the map...");
+
+  // names and geographic locations of airport codes
+  var locations = [
```
msecret: How will a new developer add these down the line? Does there need to be instructions for them?

assets/js/map.js
```diff
@@ -0,0 +1,203 @@
...
+
+  var urls = {
+    team: "/api/team/",
+    topology: "/assets/data/us-states.json"
+  };
+
+  queue()
+    .defer(d3.json, urls.team)
+    .defer(d3.json, urls.topology)
+    .await(function(error, team, topology) {
```
msecret: One comment on this async function would be to try and break it into smaller parts because its pretty big right now. So like having separate functions for different operations, which would make it more testable. I don't think this should block you though, maybe a todo comment would be good.

msecret: Also you might want to name that async function to ensure you get better traces on error.

assets/js/map.js
mbland: Soon we'll add a `_data/locations.yml` file that'll have the coordinates. We can then make this file a Jekyll-processed template that spits the location codes and coordinates into this array, and/or just pull them from the Hub's `/api/locations/api.json` endpoint.

assets/css/map.css
```diff
@@ -0,0 +1,41 @@
+svg.map {
+  display: block;
+  width: 100%;
+  min-height: 400px;
+  margin: 1em;
+}
+
+svg.map .background path {
```
msecret: Is there a reason you can't do just ".map" without the svg tag selector?

shawnbot: No, the rest of the selectors are SVG-specific, though. So I figured `svg.map` made sense for "grouping" them.

PR #99, opened 2015-02-05
Switches the content type on json files

gboone:
- Adds an entry for `api.json` as an index file on the nginx vhost (not sure if that will actually work, will try to test it locally)
- Switches the generated filename to api.json for api files

### DIFF
--- a/_plugins/api.rb
+++ b/_plugins/api.rb
@@ -174,7 +174,7 @@ module Hub
     def self.generate_endpoint(site, endpoint, title, description, data)
       return if data.empty?
       api_endpoint = File.join('api', endpoint)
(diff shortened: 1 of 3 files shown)


mbland:
_Sweeet!_ Just confirmed on my box with @gboone next to me (after setting up my commit to update the nginx running locally in vagrant). Will merge, then restart nginx on the internal Hub.

When the site redesign is ready to merge into the public Hub, I'll update the 18f.gsa.gov nginx config accordingly (or Greg can, if he wants to beat me to it!).

mbland:
Whoops, forgot to cc: @afeld @shawnbot @meiqimichelle @msecret @brethauer @elainekamlley

PR #102, opened 2015-02-06
A few adjustments to search style.

meiqimichelle:
This will work for the demo tomorrow, although I actually think that search needs to be in the nav bar eventually. It is distracting to have it at the top of every page.

Very much good enough to share functionality and get feedback :)

@mbland I may have messed something up in translation. On my local branch, when I start to search for something it clears the content of the page while searching.

screenshot 2015-02-05 21 18 30


mbland:
LGTM. I think the search bar cleared the page before, too.

PR #103, opened 2015-02-06
Adds basic subnav to team page

meiqimichelle:
This adds a basic subnav to the team page that looks like this:
screenshot 2015-02-05 21 52 53

I think is this a good solution to test with, and a good solution for the demo tomorrow.


mbland:
Cool! Definitely like where this is going.

PR #108, opened 2015-02-06
Add team map to locations page

shawnbot:
This PR (a redo of #94) introduces a team map to the `/locations/` page, and some conventions for per-page scripts and CSS. Namely, the head and scripts includes now look for `page.stylesheets` and `page.scripts`, respectively, and add the relevant `<link>` and `<script>` tags for each one. You can see this in action on the [locations page](https://github.com/18F/hub/blob/feature/team-map/pages/locations.html#L4-L9):

```yaml
scripts:
- /assets/js/vendor/d3.v3.min.js
- /assets/js/vendor/queue.v1.min.js
- /assets/js/vendor/topojson.v1.min.js
- /assets/js/map.js
stylesheets:
- /assets/css/map.css
```

[map.js](https://github.com/18F/hub/blob/feature/team-map/assets/js/map.js) then looks for `#team-map`, loads some [US states in TopoJSON](https://github.com/18F/hub/blob/feature/team-map/assets/data/us-states.json) and the team data from `/api/team/`, groups members by location (these are [hard-coded](https://github.com/18F/hub/blob/feature/team-map/assets/js/map.js#L9-L20)), and makes a map.

Each location on the map is wrapped in an `<a>` that naïvely links to `/locations/{code}/`, some of which may not exist. We could eliminate the hard-coded locations by moving their coordinates and city names into a new YAML data file, and exposing that info in the `/api/locations/` endpoint. cc: @mbland


_includes/team_members.html
```diff
@@ -4,10 +4,13 @@
       <a href="{{ site.baseurl }}/team/{{ member.name }}"><img class="team-member-thumb" src="{{ site.baseurl }}/{{ member.image }}" alt="Thumbnail photo of {{ member.full_name }}"></a>
     </div>
     <div class="team-member-info">
-      <a href="{{ site.baseurl }}/team/{{ member.name }}">{{ member.full_name }}</a>{% if member.location %} - <a href="{{ site.baseurl }}/locations/{{ member.location }}">{{ member.location }}</a><br/>{% endif %}
-      <p>{% if member.projects and member.projects.length != 0 %}
-      Projects: {% for p in member.projects %}<a href="{{ site.baseurl }}/projects/{{ p.name }}">{{ p.project }}</a>{% unless forloop.last %}, {% endunless %}{% endfor %}{% endif %}
+      <a href="{{ site.baseurl }}/team/{{ member.name }}">{{ member.full_name }}</a>{% if member.location %} -
```
msecret: <a> tags should have a title attribute for accessibility

assets/js/map.js
```diff
@@ -0,0 +1,242 @@
+(function(exports) {
+
+  // our DOM container and status message
+  var map = d3.select("#team-map")
+        .classed("loading", true),
+      status = map.append("p")
+        .html("Loading the map...");
+
+  // names and geographic locations of airport codes
+  var locations = [
```
msecret: Are there instructions for how someobody would add new locations? Could this be automated? I would imagine that if a member from a new city joins our team, we'll forget to add the new city.

assets/js/map.js
```diff
@@ -0,0 +1,242 @@
...
+    .defer(d3.json, urls.team)
+    .defer(d3.json, urls.topology)
+    .await(allDone);
+
+  function allDone(error, team, topology) {
+    map.classed("loading", false);
+    if (error) return showError(error.statusText);
+
+    // kill the status message
+    status.remove();
```
msecret: Suggestion: this function would be easier to test if status wasn't a variable out of scope for this function but was passed in.

assets/js/map.js
```diff
@@ -0,0 +1,242 @@
...
+
+    // group members by location
+    var members = d3.values(team)
+          .filter(function(d) { return d.location; }),
+        byLocation = d3.nest()
+          .key(function(d) { return d.location; })
+          .map(members);
+
+    // then assign a list of members to each location
+    locations.forEach(function(d) {
```
msecret: Suggestion, it would be a little cleaner to make this function, so var updatedLocations = locations.map , etc.

assets/js/map.js
shawnbot: I've been assuming that the `locations` list disappears once we have an API endpoint for them, and the documentation for updating that list lives as a comment in the yml file, the README, or the wiki or something. So this is a stopgap solution for now, but I'm happy to add some instructions for adding locations just in case that doesn't happen soon enough.

assets/js/map.js
shawnbot: :+1:

assets/js/map.js
shawnbot: What about replacing `.await(allDone)` with an inline function that does lines 47-51 then calls another function with just the `team` and `topology`, e.g.:

```js
.await(function allDone(error, team, topology) {
  map.classed("loading", false);
  if (error) return showError(error.statusText);

  // kill the status message
  status.remove();
  renderMap(team, topology);
});
```

assets/js/map.js
mbland: Opened #134.

mbland:
@shawnbot `View {{ member.full_name }}'s profile` is a good choice for `title`, I think. @nickbristow, any thoughts?

Merging this now, but feel free to open a new PR with the `title` attribute set. Nice work, @shawnbot and @msecret! (Now we just need to find a way to add tests... ;-)

PR #109, opened 2015-02-06
Make an org chart

shawnbot:
:construction: **WIP** :construction:

This PR introduces a new `org.yml` and `org` API endpoint that describes the 18F team structure.

First off, I could use some feedback on the YAML structure. It's a tree composed of "teams" with 18F at the root. I haven't added people to this yet, but my thinking was that we could designate team leads by adding `lead: <id>` to a team and a `members` list to each if we want to get fancy.

Then, if we decide to go down the route of putting people in here, we can talk about how best to do that in the API endpoint. I started trying to hack in a recursive function that would populate `members` lists, but I have zero Ruby experience and got hung up pretty quickly.

Next up, I'm going to work on some really basic HTML output for this on structure page.

### DIFF
--- a/_data/org.yml
+++ b/_data/org.yml
@@ -1,0 +1,24 @@
+name: 18F
+teams:
+- name: Delivery
+  teams:
+  - name: Engineering
+  - name: Product
+  - name: Practice
+  - name: Senior Product Design
+  - name: Staff & Projects
+  - name: Team Administration
+- name: Creative, Communication & Community
+  teams:
+  - name: Outreach
+  - name: Experience Design
+    teams:
+    - name: UX Design
+    - name: Visual Design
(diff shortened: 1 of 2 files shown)


mbland:
cc: @aaronsnow @quepol

Yeah, I was wondering about this; it's about time to hash it out. Going this route, it seems similar to how the `projects.yml` and `working_groups.yml` files are structured, so thumbs-up there.

In addition to this, there's the matter of the literal reporting structure. I was thinking of adding a `supervisor:` field to each team member, building the tree from the bottom up. Or, of adding a `reports:` field to supervisors, and to build the tree from the top down.

Also wondering how much of this can/should be exposed publicly, versus hiding them as `private:` properties.

PR #123, opened 2015-02-09
Restart google_auth_proxy for internal Hub

mbland:
Kept getting bitten by new user updates not taking right away; realized the `./go` script could do the trick.

cc: @gboone @afeld


gboone:
LGTM!

PR #144, opened 2015-02-11
Use prose.io as the default pages/ editor

mbland:
It's aliiiive! Let's play around with this and see how we like it, and how we can improve prose.io!

cc: @dhcole @meiqimichelle @gboone @elainekamlley @afeld

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -58,3 +58,17 @@ defaults:
       edit_info:
         repo: 18F/hub-pages-private
         prefix: pages/private/
+
+editor_url: https://hub.18f.us/edit/#
+
+prose:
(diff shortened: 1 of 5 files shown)


meiqimichelle:
Ooooooo yeeeeaaaahhh

dhcole:
Looks good. One option to cut down on a server would be to check out prose into `/edit` of this repo.

mbland:
@dhcole Hmm... as a submodule, maybe? Let me sleep on it...others welcome to weigh in too, of course!

dhcole:
Sub module or just `git clone ... prose`. Nice to see this working though!

Issue #152, opened 2015-02-13
No response after submitting search terms


esgoodman:
I started by trying to search for the "Where Am I?" content on the Hub and the search box didn't return any responses. So I tried the search box on a few pages:

https://hub.18f.us/guidelines/
https://hub.18f.us/how-to/
https://hub.18f.us/n00b

I couldn't get a search response from any of them, for any search terms.

mbland:
Dup of #119.

mbland:
Actually, I just noticed this is different; the `uglifier` introduced in #147 to compress the JavaScript seems to have broken search completely. Working on a fix.

PR #154, opened 2015-02-13
Fix minified search JavaScript using ng-annotate

mbland:
Realized after #147 that the minified `_assets/javascripts/search.js` was broken as reported in #152. After using Google Chrome's JavaScript console and researching the errors, it turns out that minifying Angular apps commonly fails due to the lack of explicit dependency injection annotations.

Fortunately, [the ng-annotate tool](https://github.com/olov/ng-annotate) fixes this problem. The changes to `_assets/javascripts/search.js` reflect running the following:

```shell
$ npm install -g ng-annotate
$ ng-annotate -ar _assets/javascripts/search.js >
  _assets/javascripts/search.js.new
$ mv _assets/javascripts/search.js{.new,}
```
Also, per the instructions for `ng-annotate`, I also added the `ng-strict-di` attribute to the search app's `div` to ensure that future changes include explicit dependency injection annotations.

Verified by inspecting the JavaScript console and typing into the search box that the error is resolved by this commit.

cc: @meiqimichelle @afeld @esgoodman

### DIFF
--- a/_assets/javascripts/search.js
+++ b/_assets/javascripts/search.js
@@ -4,13 +4,13 @@
 
 var ngHub = angular.module('hubSearch', ['LiveSearch']);
 
-ngHub.factory('pagesPromise', function($http, $q) {
+ngHub.factory('pagesPromise', ["$http", "$q", function($http, $q) {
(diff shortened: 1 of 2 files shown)


mbland:
Closes #152.

PR #172, opened 2015-02-18
Add testing cookbook link

shawnbot:
This started with @meiqimichelle suggesting that I add a link to my [VirtualBox guide](https://github.com/18F/testing-cookbook/blob/master/browser/VirtualBox.md) to the hub. But I figured that linking to the testing cookbook more generally would probably be better. Thoughts on where this goes, @mbland?


mbland:
As I talked about during the Testing Grouplet meeting today, [18F Guides](http://18f.github.io/guides/) ([repo](https://github.com/18F/guides)) would be a good place. Can you post this link there, and then update this PR to link to that instead? (Or I can do it if you prefer.)

PR #174, opened 2015-02-19
Switch import-public.rb to split file format

mbland:
cc: @afeld @gboone


mbland:
Ah, was going to let the webhooks take care of updating the public data files, but Travis needs them. Pushing an update.

PR #182, opened 2015-02-24
Export more project data via the API

mbland:
This is a step towards using the public Hub API as the data source for 18f.gsa.gov and the Dashboard.

cc: @gboone @shawnbot @afeld @meiqimichelle

### DIFF
--- a/_plugins/api.rb
+++ b/_plugins/api.rb
@@ -93,9 +93,9 @@ module Hub
 
     def self.generate_projects_endpoint(site)
       projects = site.data['projects']
-      return if !projects or projects.empty?
-      fields = ['project', 'github', 'partner', 'impact', 'stage',
-        'milestones', 'contact', 'stack', 'licenses', 'links', 'status']
+      fields = ['project', 'github', 'description', 'partner', 'partners',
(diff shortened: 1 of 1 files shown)


_plugins/api.rb
```diff
@@ -54,7 +54,7 @@ def self.generate_team_endpoint(site)
       return if team.empty?
       fields = ['first_name', 'last_name', 'full_name', 'image',
          'pif-round', 'bio', 'email', 'slack', 'location',
-         'languages', 'technologies', 'specialties']
+         'languages', 'technologies', 'specialties', 'hidden']
```
gboone: We don't have to do it now, but we can deprecate the `hidden` key whenever we want to. The front page doesn't depend on it since it now uses the existence of an image to determine whether to display the name + photo. (See: [the template code](https://github.com/18F/18f.gsa.gov/blob/staging/pages/index.html#L96) and [the `with_pic` filter](https://github.com/18F/18f.gsa.gov/blob/staging/_plugins/author.rb#L23-L33) for how we implemented that.

mbland: Ah, OK. I'll revert the commit.

Issue #134, opened 2015-02-10
Add office location API


mbland:
Once #108 is integrated, as noted in that PR, the longer-term ideal would be to replace the `locations` array in `assets/js/map.js` with a call to `/api/locations` (or `/api/locations.json`?).

cc: @shawnbot @msecret

shawnbot:
Yep, or as the URLs are currently configured, `/api/locations/api.json`. Then we'd update `map.js` to do something like this:

```js
urls.locations = '/api/locations/api.json';

queue()
  .defer(d3.json, urls.team)
  .defer(d3.json, urls.locations)
  .defer(d3.json, urls.topology)
  .await(function(error, team, locations, topology) {
    if (error) throw error;
    // join team and locations
  });
```

shawnbot:
[OpenFlights](http://openflights.org/) maintains a [CSV](http://openflights.org/data.html) of airports that includes the FAA codes and geographic coordinates we need. I used their [airport search tool](http://openflights.org/html/apsearch) to get the coordinates we're using now, but it might make sense to just grab this CSV and join it onto our data. There may be licensing issues with doing that, though.

mbland:
Very interesting idea, but yeah, it looks like licensing and/or cost may be an issue. Given the relatively few locations in which 18F has a presence, it may be more worthwhile to just maintain our own list by hand.

shawnbot:
:+1:

mbland:
Per @janearc in 18F/hub-pages-private#43:

> in team/ there are bunches of files that include a Location field; i feel like it would be useful to have a timezone, as expressed as offset from UTC (that is, US/Eastern is sometimes hard to remember, but if it's UCT-0500 it's easy to discern your delta from another team member). anyways, seems like a reasonably good thing to have, and since it looks like everyone is using airport codes, it should be easy to programmatically apply?

Per me:

> Great idea! We need to establish an independent list of location codes anyway, to merge lat/longs into the data consumed by the locations map. Would be painless to include this as well.

shawnbot:
Yes! Are we ready to add coordinates to the YAML, @mbland?

mbland:
Yep! Was hoping to get to that today, but looking like tomorrow at the earliest for me. Feel free to try beating me to it. (Also, note that Becky changed her `location:` from `BOS` to `BLD`.)

shawnbot:
OK, I'll take a look!

janearc:
[...]

PR #186, opened 2015-02-25
Use locations endpoint, nicer tooltips

shawnbot:
:construction: **Not ready for primetime!** :construction: 

I've brought in [d3-tip] for nicer tooltips and integrated the locations API. However, I started this work a while ago and so it's out of step with master, and when I merged master into this branch I ended up with a very conflicted `pages/locations.html` since I modified the template to properly pluralize "member(s)". I could use some help with the merge.

After getting master merged in, I'd like to move the geographic coordinates into the locations API endpoint, but there's currently no `locations.yml` to move them into. @mbland, is that still in progress?

[d3-tip]: https://github.com/Caged/d3-tip


mbland:
Well, regarding a `locations.yml`, #134 is still open. :-)

assets/css/map.css
```diff
@@ -38,3 +38,48 @@ svg.map .tip text {
...
+    height: 0;
+    width: 0;
+    position: absolute;
+    pointer-events: none;
+    border-color: rgba(0, 0, 0, 0);
+    border-right-color: #000;
+    border-width: 7px;
+    margin-top: -7px;
+  }
+
```
afeld: FYI, Jekyll supports SASS.

assets/js/vendor/d3-tip.js
```diff
@@ -0,0 +1,299 @@
+// d3.tip
```
afeld: How about adding via Bower?

assets/css/map.css
shawnbot: We don't *have* to use it, though, right? That's one fewer file to process during the build!

afeld: Sure – just that what you wrote could be simplified.

assets/js/vendor/d3-tip.js
afeld: We can deal with this separately... we should probably take a more holistic look at how to handle JS dependencies.

PR #187, opened 2015-02-25
Export location latitude, longitude, and timezone

mbland:
Pertains to #134. After this, the location map Javascript can use the API to pull in locations, lat/lngs.

cc: @afeld @shawnbot @msecret @janearc

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -103,7 +103,7 @@ GEM
     sprockets-sass (1.3.1)
       sprockets (~> 2.0)
       tilt (~> 1.1)
-    team_hub (0.0.4)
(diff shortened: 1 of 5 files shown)


afeld:
May want to add a note somewhere that the `locations.yml` may need to be updated if we get a new teammate in a new location? Otherwise :shipit:

mbland:
@afeld Good catch. Travis is really hungover today, but will wait to merge once it catches a second wind. `./go ci_build` passes locally, at least, so I expect it'll be OK.

Also, should've looped @dhcole in earlier.

mbland:
Gah, giving up on Travis for now. Merging.

PR #188, opened 2015-02-25
Get locations from Hub API; only show if has team

mbland:
Look, Ma! I can does Javascript!

cc: @shawnbot @msecret @afeld @dhcole

### DIFF
--- a/assets/js/map.js
+++ b/assets/js/map.js
@@ -6,24 +6,10 @@
       status = map.append("p")
         .html("Loading the map...");
 
-  // names and geographic locations of airport codes
-  // you can find lat/longs here: <http://openflights.org/html/apsearch>
-  var airports = [
-    {code: "DCA", label: "Washington", location: [-77.037722, 38.852083]},
-    {code: "SFO", label: "San Francisco", location: [-122.374889, 37.618972]},
-    {code: "CHI", label: "Chicago", location: [-87.631667, 41.883611]},
-    {code: "DAY", label: "Dayton", location: [-84.219375, 39.902375]},
-    {code: "DEN", label: "Denver", location: [-104.673178, 39.861656]},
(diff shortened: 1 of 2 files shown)


assets/js/map.js
```diff
@@ -108,7 +99,7 @@
   function renderMapLocations(locations) {
     // size accessor (keeping things DRY)
     var size = function getSize(d) {
```
afeld: @shawnbot Why does this method have two names?

afeld:
SO PROUD. :shipit:

assets/js/map.js
mbland: I don't understand. Two names?

afeld: JavaScript is, um, :sparkles: fun :sparkles: because you can define methods two ways:

```javascript
var getSize = function(d) {
  // ...
};

// or

function getSize(d) {
  // ...
}
```

Shouldn't need the name on the declaration side *and* the assignment side.

mbland: Ah. I'll let @shawnbot answer to that. ;-)

shawnbot: The reason is to make debugging easier. Anonymous functions will (well, *could*—v8 has gotten better about this) be listed as anonymous in a stack trace if something goes wrong. The tradeoff is this: [function hoisting] means that your function definition may be moved to the top of the scope, but you can avoid it by assigning a function in a `var` statement. So if you want to avoid hoisting *and* see the name of your function in a stack trace, you have to do this.

[function hoisting]: http://elegantcode.com/2011/03/24/basic-javascript-part-12-function-hoisting/

afeld: o_O

Wow. I've never heard that argument before, or seen it done.

mbland: Hey @shawnbot, that might be something worth mentioning in the [Testing Cookbook](http://18f.github.io/testing-cookbook) somewhere.

PR #189, opened 2015-02-25
Deploy public hub, including 20150223 snippets, updated location and project APIs

mbland:
cc: @gboone @OriHoffer @afeld


afeld:
:shipit:

PR #192, opened 2015-03-02
Export 18F team members as part of /api/pif_team/

mbland:
@gboone This mixes in the PIF history of current 18F team members.


gboone:
Sweet, I'll wait on travis to do it's thing.

PR #205, opened 2015-03-05
initialize submodules as part of setup

afeld:
Couldn't figure out why the private pages weren't displaying locally!

### DIFF
--- a/go
+++ b/go
@@ -58,6 +58,7 @@ def init
     puts "Bundler installed; installing gems"
   end
   exec_cmd 'bundle install'
+  exec_cmd 'git submodule update --init'
 end
 
 def update_components


mbland:
I explicitly _don't_ want to do this; users who don't have access to our repos will get an error. It's a small extra step documented here: https://github.com/18F/hub/blob/master/pages/101/go.md#initializing-the-environment

Though, if you can think of a way to gracefully handle errors and notify the user that submodules _weren't_ initialized, give it a shot!

parkr:
If you just do `system("git submodule update --init")`, it won't exit. Looks like `exec_cmd` will fail if the command fails, but uses `Kernel#system` under the hood:

```ruby
 def exec_cmd(cmd)
   exit $?.exitstatus unless system(cmd)
 end
```

PR #208, opened 2015-03-06
Fix authenticated emails for jekyll 3.0.0.beta2

mbland:
Seems the Liquid used by 3.0.0.beta2 doesn't add as many newlines as before.

cc: @afeld @adelevie


_layouts/hub-authenticated-emails.txt
```diff
@@ -1,2 +1 @@
-{% capture addrs %}{% for i in page.addrs %}{{ i }}{% unless forloop.last %}
-{% endunless %}{% endfor %}{% endcapture %}{{ addrs }}
+{{ page.addrs }}
```
afeld: Why not keep the loop in the template, but just add a newline within the loop?

mbland: Because the loop was adding newlines before the move to 3.0.0.beta2, and now it's not adding any at all, putting all of the email addresses on one line with no spaces in between.

Issue #190, opened 2015-02-27
Use Midas Profiles for Hub


dhcole:
Per conversations with @ultrasaurus @meiqimichelle @mbland, we're looking into using Midas user profiles to manage our team profiles on the Hub. This ticket should capture our approach for integration. Please take a look and let me know if I missed anything.

Midas has an interface for allowing users to store data about their location, skills, and topics of interest. Midas has a simple to use interface for managing this information.

![screen shot 2015-02-27 at 5 29 20 pm](https://cloud.githubusercontent.com/assets/170641/6422405/339d1ae2-bea6-11e4-91b1-1ec3c0146179.png)

Here are the steps for a clean integration with the hub:

- [ ] Set up MyUSA authentication on Midas (Could also move the hub to use MyUSA instead of Google Auth proxy)
- [ ] Expose profile data though Mida's JSON API
- [ ] Maintain list of 18F team members' email addresses and office assignment in hub `team.yml`
- [ ] Adjust pages that list team profiles, skills, locations, etc to use client-side JS requests from JSON API (this could be the Midas API or a local JSON API for hub instances that don't use Midas, so it's not a hard dependency)
- [ ] Get the team to update their Midas (Open Opportunities profiles

I'll be working on the Midas-related tasks toward the beginning of next week and move over to adjusting hub templates toward the second half of the week.

dhcole:
Refs #134 and #97

mbland:
Super-excited about this! Also wanna pull in @afeld and @shawnbot at the very least.

And @yozlet, would love to grab some time to play with getting the Hub set up with MyUSA.

Issue #89, opened 2015-02-03
MyUSA Authentication


mbland:
In order to potentially pitch the Hub to groups that don't use Google-managed domains, we need to investigate how we could potentially use MyUSA for Hub authentication.

cc: @yozlet @afeld @dlapiduz @cpapazian

cpapazian:
You don't need the googles. An email address is sufficient.

afeld:
Sounds like https://github.com/bitly/google_auth_proxy might need to be forked...

cpapazian:
@afeld not sure i follow.

cpapazian:
@mbland MyUSA does not require an OAuth integration. We have a passwordless authentication workflow that sends a single-use token to your email address.

dlapiduz:
@cpapazian `google_auth_proxy` is an app that only lets traffic through if you are authenticated with google apps. I believe @afeld is saying that we should fork it and use MyUSA instead of G+.

+ :100: :+1:

afeld:
'zactly. More info: https://github.com/18F/hub/blob/master/deploy/README.md#google-auth-proxy

afeld:
Should also note that @GUI doesn't have access to our private Hub at the moment, because he doesn't have a GSA.gov email.

mbland:
Ask, and ye shall receive! Behold: (the changes needed to adapt the google_auth_proxy to MyUSA)[https://github.com/bitly/google_auth_proxy/compare/master...18F:myusa-staging?expand=1]. You can try it out here: https://hub.18f.us/myusa/

Next steps:
- [ ] Q: Should we rename the fork and keep it MyUSA specific, or try to genericize and commit upstream?
- [ ] Either get all 18Fers and guests onto MyUSA, or update the g_a_p to perform multiple authentication methods. (Or both.)

dlapiduz:
@mbland this is great!

I would fork it, rename and do MyUSA specific indeed. I think that the point of the g_a_p is to keep things extremely simple.

IMHO we should have everyone in MyUSA. Dogfooding FTW

adelevie:
I agree with @dlapiduz. g_a_p isn't a Golang flavor of Omniauth or Passport. It does one thing, simply. Same should apply to a MyUSA variant.

[...]

PR #211, opened 2015-03-06
Use Midas API to supplement team profile data

dhcole:
Implements the changes outlined in #190.

Adds data from Midas profiles to team member profiles. Uses a client-side template to show a member's skills and interests, which are relabeled / simplified from languages, technologies, and specialties.

- [x] Merge "languages" and "technologies" into "skills" and rename "specialties" to "interests"
- [x] Request data from Midas's API on team profile pages matching on member email address
- [x] Merge skills and interests from Midas with those on team profile
- [x] New client-side skills and interests landing pages
- [x] Update links on teams landing page
- [x] Set Midas URL and API token

cc @ultrasaurus @mbland


dhcole:
@mbland This is working now. Would love your eyes on it when you have a chance. Ping me on Slack and I'll send you a key to test with.

To ship this, we need to set the Midas API token and switch the Midas URL from dev to production.

How are you setting the keys for `google-auth-proxy`? Can we use the same process to set this key?

mbland:
@dhcole Will try to loop back with you on this this week. The `google_auth_proxy` key is set in a config file that the `google_auth_proxy` reads on startup. I'm not sure yet if/how to pass tokens along from the `google_auth_proxy`, but I imagine it shouldn't be difficult to do.

FYI, I hacked the `google_auth_proxy` to use MyUSA per #89. More work is needed, but the proof-of-concept is sound. (And now I'm rapidly becoming acquainted with OAuth2 API conventions and reverse proxies.)

ultrasaurus:
FYI: we updated the [production open opps](https://midas.18f.us) to include MyUSA authentication --- what needs to happen for this work to be merged in?

mbland:
Just a merge from master (since GitHub says it can't be automatically merged) and @dhcole's heads-up that he's using the auth token included as part of #241.

dhcole:
@mbland Just resolved the conflicts. There's one more issue I could use your help with:

I set up HTML pages for `/skills` and `/interests` that use the client-side data aggregation so they can pick up Midas data. However these pages are inaccessible because a plugin is generating them automatically. What's the best way to resolve this?

mbland:
Comment out `Skills.generate_pages(site)` in `_plugins/hub.rb`.

dhcole:
Cool, will give that a shot

dhcole:
@mbland Thanks, looks like that works well.

This should be ready to :ship:

_includes/head.html
```diff
@@ -90,7 +90,7 @@
 // export some helpful site-wide variables
 window.site = {
   baseurl: "{{ site.baseurl }}"{% unless site.public %},
-  access_token: "<!--# echo var="access_token" -->"{% endunless %}
+  access_token: '<!--# echo var="access_token" -->'{% endunless %}
```
mbland: Why do these quotes need changed?

mbland:
Can you post a couple screenshots of the new versions? Just for completeness's sake.

_includes/head.html
mbland: Ah, I think I see why you changed them. But the Server-Side Include replaces `var="access_token"`, so there's no issue with using `"` in both places. Don't bother changing it back, though.

dhcole: Using double quotes, the string is broken by the double quotes are `"access_token"`. This makes the JS invalid and so it fails linting.

Also, when running the hub locally without nginx, this server side include doesn't work, so the broken string causes a JS error that blocks the rest of the script.

By switching to single quotes, everything stays valid. If there SSI doesn't get replaced, the string is just sent as the token, which is invalid, and midas authentication fails (which is expected and OK behavior).

mbland:
Hmm, I pulled this branch to run it locally, and while my individual team page appears OK, the pages for individual skills appear to be blank.

mbland:
D'oh! PEBKAC...need to run in Vagrant, not with `./go serve`...

mbland:
...and/or hand-jam my own access token in temporarily...

dhcole:
One note: pages for individual skills get routed to the sample root skill page `/skills/?javascript` instead of `/skills/javascript`

Otherwise, the templates should look the same as before with the addition of data from Midas profiles. Here are some screenshots.

`/skills/`
screen shot 2015-04-17 at 6 02 06 pm

`/skills/?javascript`
screen shot 2015-04-17 at 6 01 57 pm

`/team/dhcole/`
screen shot 2015-04-17 at 6 01 32 pm

dhcole:
My profile above has some skills showing from `_data` and some from Midas, to show how they're merged into the same template.

mbland:
For the folks tuning in at home: I'm having trouble reproducing Dave's screenshots locally. My access_token is valid, and I've authorized Midas to access my email address. There is a curious stack trace I'm seeing locally that implies an interference with search somehow (which doesn't appear on the current production Hub; I commented out the `js_compressor:` line in `_config.yml` to get proper names in the stack trace):

screen shot 2015-04-17 at 7 03 06 pm

Dave and I'll get together on Monday to see if we can't get to the bottom of why.

mbland:
@dhcole I'll be kiss my... We need to remove `Skills.generate_pages(site)` from `_plugins/hub.rb`. Maybe just remove `_plugins/skills.rb` altogether, too.

dhcole:
Ah, I pushed that to the wrong branch (origin vs dhcole). Here's the commit that comments it out. Does that work for you? b075983e4302e9b80404fd5212a359b143010e2e

dhcole:
Hey @mbland just checking if the change in https://github.com/18F/hub/pull/211/files#diff-2739ce3f9a0195fafcfad4bfe2a468df cleared up the issue you saw.

mbland:
@dhcole Yes, it did. Only thing is, the performance kind of worried me; DM'd you in Slack a few days ago about that. Happy to pick this up tomorrow.

mbland:
Err, tomorrow's tight...but Thursday/Friday for sure.

dhcole:
@mbland The performance for this change should only be an additional ajax request and some (minimal) theming. Testing against the production-hosted Midas instance, I see the following loading time:

screen shot 2015-04-29 at 4 18 30 pm

With a refactor that's live on staging and about to go live on production, this endpoint will also be twice as fast:

screen shot 2015-04-29 at 4 19 45 pm

So the overall impact of this PR should be ~250ms on pages that pull in data from Midas.

However, I am seeing an additional +1000ms delay due to something happening in the search.js code.

screen shot 2015-04-29 at 4 17 16 pm

screen shot 2015-04-29 at 4 13 53 pm

I'm not sure what would be causing this, or whether it's related. One possibility is that this PR stores member data as json within the page's HTML. It's possible the lunr script is trying to index that additional data.

If that's it, is there a way to have lunr skip <script> tags? cc @afeld

mbland:
@dhcole It seems that `require.js` is triggering an `onScriptLoad` event that ultimately invokes a rebuild of the lunr index. I'm going to poke at building the index server-side for now, per olivernn/lunr.js#60, to see if that'll fix the load time issue altogether.

However, I'd also like the page to contain links _before_ the Midas call to update the page, so that there's info to index in the first place. Plus, [progressive enhancement](http://en.wikipedia.org/wiki/Progressive_enhancement) and all that. Any chance you can tackle that in parallel? That is to say, the server-side data is included in a `div`, that is then swapped-out with the Midas-generated `div`? Let me know if what I'm asking makes sense.

mbland:
#278 definitely speeds up the rendering _a lot_. Now the index is loaded in background chunks rather than being compiled locally after the `onScriptLoad` event. And on a proper webserver that returns 304s, it shouldn't really be a factor at all.

Still, what do you think about my progressive enhancement issue, @dhcole?

dhcole:
Great to hear that #278 speeds things up. The delay we're talking about now in the 100-200ms level is negligible given you can't even scroll the page that fast. I do think getting this data into the search index is valuable, but I'm not sure it's a big deal at first.

It will require some refactoring render server- and client-side and likely delay getting this done for another day or into next week. Up to you whether you want that to block merging in this work.

I will make time for it this afternoon and see how far I can get.

mbland:
What if instead of (or in addition to) the current Liquid template rendering the data as HTML, we rendered it as JSON (instead of your current template), and then merged the Midas data into that? Would that work, or take a lot of work?

I'll think on the search bits. I am eager to get this done by tomorrow afternoon at the latest.

mbland:
#278 is in. Merge and give it a shot?

dhcole:
Cool. Wil try this with an attempt at a server template tomorrow morning.

mbland:
Per a private DM with @dhcole, we're going to follow up with a later PR to better integrate render-time skills/interests from the YAML files with Midas-supplied data. But finally, at long last, we're ready to launch Midas+Hub integration for the masses.

Thanks for hanging in there and seeing it through, @dhcole and @ultrasaurus.

PR #221, opened 2015-03-17
add 18F-specific contribution instructions

bsweger:
Added directions for contributing to the Hub as an 18F team member, to make it clear that we don't have to fork. Are these git instructions for handling the submodule the right way to do that?

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ work with the Hub development environment.
 
 ### Contributing
 
-1.  Fork the repo ( https://github.com/18F/hub/fork ). If you're an 18F team member, you'll likely find it easier to clone the repo instead of forking it (`git clone --recursive git@github.com:18F/hub.git`). The recursive clone ensures that you'll grab the contents of the private data submodule.
+1.  Fork the repo ( https://github.com/18F/hub/fork ). If you're an 18F team member, you'll likely find it easier to clone the repo instead of forking it (`git clone --recursive git@github.com:18F/hub.git`). The recursive clone ensures that you'll grab the contents of private submodules.
 2. Create your feature branch (`git checkout -b my-new-feature`)
 3. Commit your changes (`git commit -am 'Add some feature'`)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,6 +61,19 @@ Feel free to ping [@mbland](https://github.com/mbland) with any questions you
 may have, especially if the current documentation should've addressed your
 needs, but didn't.
 
+### Contributing as an 18F Team Member
+
+1. Clone the repo, including the submodule for private data ( `git clone --recursive git@github.com:18F/hub.git` )
```
mbland: Since this is the only step that's different, I'd just make this a sub-item in the earlier list. (And I didn't know about `git clone --recursive`. Nice!)

bsweger:
Good point @mbland. Is this better?

README.md
```diff
@@ -51,7 +51,7 @@ work with the Hub development environment.
 
 ### Contributing
 
-1. Fork the repo ( https://github.com/18F/hub/fork )
+1.  Fork the repo ( https://github.com/18F/hub/fork ). If you're an 18F team member, you'll likely find it easier to clone the repo instead of forking it (`git clone --recursive git@github.com:18F/hub.git`). The recursive clone ensures that you'll grab the contents of the private data submodule.
```
mbland: I'd say "private submodules" instead of "private data submodule". There's `_data/private` and `pages/private`.

README.md
```diff
@@ -61,6 +61,11 @@ Feel free to ping [@mbland](https://github.com/mbland) with any questions you
 may have, especially if the current documentation should've addressed your
 needs, but didn't.
 
+Feel free to ping [@mbland](https://github.com/mbland) with any questions you
```
mbland: This duplicate paragraph should be removed.

bsweger:
Doh--copy and paste fail. Thanks for the catch and the clearer working suggetion, @mbland

mbland:
Thanks for the really helpful update, Becky!

PR #222, opened 2015-03-17
Update to use Midas-based skills categories

mbland:
Once this goes in, I'll merge the `data-private` changes into the `data-private` master branch.

cc: @dhcole @meiqimichelle @afeld @juliaelman @gboone


mbland:
BTW, I can add my `munge.rb` script if that'd be helpful for any reason.

mbland:
Whoops, meant that for 18F/data-private#68.

dhcole:
@mbland looks great! -- should this be updated too? https://github.com/18F/hub/blob/98e0c918382884dea288c264a2fcbbb76dfb82dc/_includes/nav.html#L21-L23

Also, there's this, but it will be updated as part of my PR for using midas data (which can go now -- it just won't pull in any midas data until we get the `access_token` in place.

https://github.com/18F/hub/blob/41bb2151649d0ae03ab12112b4c8cdb3bb4df35b/_layouts/team_member.html#L24-L35

mbland:
Good catches, @dhcole. The `nav.html` file was vestigal, and I've updated the team member layout. PTAL?

dhcole:
Looks great! :+1:

PR #223, opened 2015-03-18
Add skills and interests to the team API

mbland:
Missed this in #222. cc: @dhcole


dhcole:
@mbland do we need to update this line too? https://github.com/18F/hub/blob/skills-interests-api/_plugins/api.rb#L148

mbland:
Done.

Why the hell do I keep missing these things? No more:
```sh
$ find . -name '*.rb' | xargs egrep '([Ll]anguages|[Tt]echnologies|[Ss]pecialties)'
```

Did the same for `.html` and `.md`. Think I got 'em all.

dhcole:
:+1: Looks great! Will merge after tests pass.

Issue #213, opened 2015-03-10
Reimagine the design of the Hub


meiqimichelle:
If there is a designer with spare cycles, I would be happy to re-imagine the Hub design again now that we have learned more about where we're going and what people what to use it for. Ping me if you want background (sorry that I don't have time to write a nice issue for you now!) :(

meiqimichelle:
Some issues:
- now that we've learned how people are using the site, we have outgrown the current site nav. Re-assess content structure.
- think about ways to facilitate browsing through the tons of info. 'How-to' may not be the way to go.

mbland:
- [ ] Ensure every page has a `title:` in its front matter.
- [ ] Add `tags:` to front matter where appropriate.
- [ ] Create a tool/plugin for discovering missing titles (and tags?).
- [ ] Follow-through on @jenniferthibault's design (implicated in #269).
- [ ] Draft content-first design a la Stephanie Hay (h/t @brethauer)

meiqimichelle:
Perhaps we should move the first three of those tasks to another issue. They don't relate to re-imagining the design of the Hub. What do you think @mbland ?

mbland:
@meiqimichelle Makes sense, though #274 takes care of the first and third. Checked 'em off here, but will create a new issue.

meiqimichelle:
:+1:

PR #228, opened 2015-03-19
Add Working Groups nav bar item

mbland:
@meiqimichelle: Folks have been asking more frequently "How many working groups are there and what are they?" Happy to entertain a different means of surfacing them, but here's my stab.
2015-03-19-hub-wg

### DIFF
--- a/_includes/nav-main.html
+++ b/_includes/nav-main.html
@@ -5,6 +5,7 @@
     </a>
     <a href="{{ site.baseurl }}/team"><li>Team</li></a>
     <a href="{{ site.baseurl }}/projects"><li>Projects</li></a>
+    <a href="{{ site.baseurl }}/wg"><li>Working Groups</li></a>
     <a href="{{ site.baseurl }}/guidelines"><li>Guidelines</li></a>
     {% unless site.public %}
     <a href="{{ site.baseurl }}/structure"><li>18F Structure</li></a>


meiqimichelle:
I'm not sure of the best solution yet, but I'd like to wait to make changes until we have a designer (and/or anyone who has time to volunteer) (maybe a fixit task for the new folks?) take a look at the site nav again overall. For example, if we make a top-level nav item for Working Groups, then my next question is...."but where are the Guilds?" and I don't want to go down the path of making All The Things top-level links.

Is there an issue for this, erm, issue? ;)

mbland:
`/wg/` contains both working groups and guilds. In the short-term, I'd prefer to have some kind of link to that page, even if there's a better way to do it later. Right now, it's not directly accessible from anywhere.

Would you prefer to file a site-nav issue? I can do it, but I doubt I'd be able to express the concerns as well as you. ;-)

And yes, it indeed might make a great fixit task!

meiqimichelle:
Right now, working groups and guilds are under "18F Stucture", right? That is how they're accessible?

meiqimichelle:
Also, I added some detail to #213 to address the site nav. It was on my mind when I wrote that issue.

mbland:
OK, I totally missed the "18F Structure" listing... Maybe that'll do. But yes, we probably do need to have some kind of site-structure pow-wow next week.

PR #231, opened 2015-03-23
Use team_hub v0.0.6 to copy private assets

mbland:
This function (and its test) is based on `TeamHub::PrivateAssets::copy_directory_to_site`. Once this goes in, I'll replace that function with this one, and replace all calls to use this new function.

cc: @afeld @melodykramer @bsweger


afeld:
Per https://github.com/18F/team_hub/pull/9#issue-64052353, will wait for the update.

mbland:
Updated and renamed the PR; just need to wait for Travis.

PR #234, opened 2015-03-25
Added coveralls.io badge to README

arowla:


mbland:
The Travis breakage here was unrelated; fixed in #235.

PR #235, opened 2015-03-26
use new Jekyll 3 beta numbering scheme

afeld:
I guess they changed their version naming scheme?

https://rubygems.org/gems/jekyll

Same Jekyll code, but updated some other gems. Fixes the build.


mbland:
D'oh! Thanks for the catch, @afeld.

PR #241, opened 2015-04-03
Embed access token as Javascript variable via SSI

mbland:
The motivation for this is to pass the authenticated user's MyUSA access token
through so that Javascript can be used to request Midas data per #211. More
generally, this should allow us to integrate other backend services that
accept MyUSA access tokens. (Care must be taken, of course, to ensure that
tokens are sent only over secure connetions.)

The upstream auth proxy will pass the OAuth `access_token` into the Hub server
via the `X-Forwarded-Access-Token` HTTP header. We then use the Nginx `set`
directive to rewrite this header's value as the `access_token` variable, which
then gets assigned to a Javascript variable via the `#echo` Server-Side
Include directive.

cc: @dhcole @afeld @adelevie @yozlet @jackiekazil

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -89,7 +89,8 @@
 <script>
 // export some helpful site-wide variables
 window.site = {
-  baseurl: "{{ site.baseurl }}"
+  baseurl: "{{ site.baseurl }}"{% unless site.public %},
+  access_token: "<!--# echo var="access_token" -->"{% endunless %}
 };
 </script>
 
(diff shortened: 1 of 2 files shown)


mbland:
@dhcole This good to go, now that bitly/google_auth_proxy#80 is in and running on hub.18f.us. After merging this, then we should then be able to get #211 in soon.

dhcole:
Nice work. Great that this made it upstream. Keeps this code very simple. :+1: on this, and I'll review #211 and update accordingly.

PR #246, opened 2015-04-07
Add Julia Elman to 18F front end guild.

juliaelman:


mbland:
Could you add this in 18F/data-private instead?

Issue #248, opened 2015-04-08
Prose.io confusion


jenniferthibault:
n00b user story: I didn't know what prose was, and when I landed on a prose page by clicking "Edit this page," I didn't know where I had been taken (it clearly wasn't GitHub), what was happening, if I was in the right place to edit, or if the changes I made were automatically committed, or if they'd go somewhere to be approved.  I expected to go straight to GitHub when I clicked edit.

mbland:
Yeah, going straight to GH is what that link used to do. :-) Lemme see what I can do...

gbinal:
+1 to sticking with linking straight to github.com

meiqimichelle:
Can this be closed @mbland ?

PR #249, opened 2015-04-08
Link to instructions under "Edit this page" link

mbland:
Attempts to address #248:

edit-instructions-link

cc: @jenniferthibault @meiqimichelle @melodykramer @bsweger


meiqimichelle:
You know what I'll say (if we need a FAQ/instructions, we're doin' it wrong). But if you are suggesting this as a short-term band-aid until we address the larger usability issue...??

mbland:
Well, I'm all ears if we can take a bite of out the usability issue sooner than later. :-) Perhaps we should drag @dhcole in here, too.

PR #251, opened 2015-04-09
Update OpenFEC team

arowla:


mbland:
Hey @arowla, can you do this in the data-private repo instead?

PR #259, opened 2015-04-17
simplify 4/17

bsweger:
Too little, too late, and if I had to do it again, I'd simplify or eliminate the entire structure of pages around the doc sprint, but at the very least let's remove the bunch of text around today's event and add a link to a very simple google doc in case we want to capture what we'll be working on.

The to-do list isn't great, so feel free to add to it, if you'd like. At this point, thought, I think we're all just planning to work through our own Hub-related punch lists.

### DIFF
--- a/pages/docsprint/onboarding.md
+++ b/pages/docsprint/onboarding.md
@@ -1,87 +1,16 @@
 ---
 permalink: /docsprint/onboarding/
 published: true
-title: Untitled
+title: Onboarding Doc Sprint
 ---
 
 # Onboarding Doc Sprint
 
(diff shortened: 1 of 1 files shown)


mbland:
I actually _don't_ want to delete all of this. Perhaps just reorganize it a bit so the big ask is up top (the tasks + the form). Maybe even split into two or three pages. But I really, really like your explanations and want to edit and build on them, not toss them out.

bsweger:
I want to toss it! I don't think we need any more pages and text related to doc sprints. Spring cleaning!

PR #260, opened 2015-04-17
load search code asynchronously

afeld:
Takes the `Load` time reported in the Chrome Dev Tools from 2.66s to 943ms.


mbland:
Nice!

Issue #169, opened 2015-02-18
Improve build slowness


mbland:
Per yesterday's Documentation Working Group discussion, build times for the Hub appear to be creeping up. This makes Javascript development especially painful.

Using the `ruby-prof` gem via the following command reaffirmed common knowledge that Liquid accounts for most of the time:

```shell
$ bundle exec ruby-prof --printer=call_stack --file=stack.out /usr/local/var/rbenv/versions/2.2.0/bin/jekyll -- build
```

Cloning the Jekyll repo and building a 3.0.0.pre.beta1 gem from the `master` branch didn't appear to impact the time very much.

Some relevant links:
- jekyll/jekyll#3009
- jekyll/jekyll#3116
- jekyll/jekyll#3324
- jekyll/jekyll#3350
- https://tuananh.org/2014/07/26/speed-up-jekyll-site-generation-with-rsync/

Possible approaches:
- Reduce the number of Liquid expressions in Hub templates
- Reduce number of Hub layouts
- Optimize Jekyll
- Optimize Liquid
- Come up with an alternate Javascript development scheme

cc: @shawnbot @msecret @afeld @gboone

dhcole:
In case this is helpful: https://github.com/jekyll/jekyll/issues/1140#issuecomment-25559506

afeld:
Made an issue for jekyll_pages_api as well: https://github.com/18F/jekyll_pages_api/issues/20

envygeeks:
Are these raw benches or raw + parboiled benches?  We have incremental regeneration so I would be interested in seeing what it's like after the raw build (on 3.0) but if that does not speed things up then we do have a problem that needs to be solved.

mbland:
I'm sure I didn't do a very thorough job with the benchmarks, as I'm not much of a Rubyist. If you've a preferred methodology documented somewhere, I can study up and take another shot. Anecdotally, it seemed like rebuilds were still on the order of 5-6s on my Yosemite machine, even with 3.0.0.pre.beta1; @shawnbot, running Mavericks, was getting O(30s) using 2.5.3.

FWIW, we're also considering changing the architecture of the application in a way that may mitigate the problem (though that's not the only reason for the architectural change).

ac70:
this may or may not help, but I cloned my jekyll repo and started setting it up to use with Grunt.   I didn't quite get there, but when I went back to trying to serve with Jekyll it was taking 1m+ for every save.  Maybe the added files caused this?

[...]

PR #261, opened 2015-04-17
Replace edit_link include with custom Liquid tag

mbland:
Part of the attempt to address generation speed per #169. Not sure how much it helps; @afeld could you perhaps pull the branch and see what happens on your machine?

cc: @gboone


afeld:
My build times are all over the place, on both this branch and `master`. Seeing everything from 5-24 seconds :unamused: Tested with both `touch assets/js/search.js` and `touch _includes/scripts.html`.

PR #269, opened 2015-04-27
Approximate Jen Thibault's Hub index page concept

mbland:
Here's @jenniferthibault's original concept:

hub landing private-v1

Here's my current approximation, implemented in this PR (minus the login image):

hub-redesign

@meiqimichelle and @shawnbot: Care to help me whip this into shape?

PR #272, opened 2015-04-29
removing standards from guidelines pages

NoahKunin:
Standards are different from guidelines. I think we'll re-address this as part of the redesign in #269


mbland:
Hey @NoahKunin, it seems like we should have these links transferred somewhere else in the Hub than removed entirely. Mind adding a `standards.md` page to 18F/hub-pages-private? Or I can do it, and merge this when that's done.

PR #274, opened 2015-04-29
Add tests for untitled posts.

monfresh:
I added two new tests that check for empty post titles and titles that are literally set to "Untitled". The failure error will show you an array of URLs for posts that don't have titles.

I also added the coverage folder to `.gitignore` since it doesn't look like it's in the GitHub repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ _site_public
 .vagrant
 _test/tmp
 bower_components
+coverage/*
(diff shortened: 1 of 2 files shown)


monfresh:
Obviously, Travis will fail until you add titles to the posts :)

mbland:
ZOMG, this is brilliant! Thanks, @monfresh! I'm going to hold off on merging until a little later this afternoon, so's I can get those title-less pages fixed right after this goes in.

PR #278, opened 2015-04-30
Build the lunr.js index server-side

mbland:
Should help with the slowness in #211. Will try it momentarily to see.

If there's a better or prettier way to do this, let me know. Testing ideas also welcome. (Will try to think on this this afternoon.)

cc: @dhcole @afeld

PR #279, opened 2015-05-01
Ship a url-to-doc map for search, not pages.json

mbland:
Size comparison:
$ ls -l _site/api/v1/pages.json _site/url_to_doc.json
-rw-r--r--  1 _  _  675772 May  1 15:34 _site/api/v1/pages.json
-rw-r--r--  1 _  _   50932 May  1 15:34 _site/url_to_doc.json

$ python -c 'print "%0.2f%%" % ((50932 / 675772.0) * 100)'
7.54%

cc: @afeld @dhcole

### DIFF
--- a/_plugins/search.rb
+++ b/_plugins/search.rb
@@ -11,14 +11,27 @@ module Hub
         'assets', 'js', 'vendor', 'lunr.js', 'lunr.js'))
       cxt.load(File.join(site.source, '_plugins', 'search.js'))
       cxt.eval("var corpus = #{corpus_page.content};")
-      cxt.eval('corpus.entries.forEach(function(page) {index.add(page);});')
+      cxt.eval("var url_to_doc = {};")
+      cxt.eval(
+        'corpus.entries.forEach(function(page) {' +
+        '  index.add(page);' +
+        '  url_to_doc[page.url] = {url: page.url, title: page.title};' +
+        '});')
 
       index_page = JekyllPagesApi::PageWithoutAFile.new(
         site, site.source, '', 'index.json')
       index_page.content = cxt.eval("JSON.stringify(index.toJSON());")
-      index_page.data['layout'] = nil
-      index_page.render(Hash.new, site.site_payload)
-      index_page
+
+      url_to_doc_page = JekyllPagesApi::PageWithoutAFile.new(
+        site, site.source, '', 'url_to_doc.json')
+      url_to_doc_page.content = cxt.eval("JSON.stringify(url_to_doc);")
+
+      pages = [index_page, url_to_doc_page]
+      pages.each do |page|
+        page.data['layout'] = nil
+        page.render(Hash.new, site.site_payload)
+      end
+      return pages
     end
 
     def self.find_corpus_page(pages)
--- a/_plugins/search.rb
+++ b/_plugins/search.rb
@@ -4,19 +4,13 @@ module Hub
   class SearchIndexBuilder
     def self.build_index(site)
       corpus_page = find_corpus_page(site.pages)
-      return nil if corpus_page == nil
+      raise 'Pages API corpus not found' if corpus_page == nil
 
       cxt = V8::Context.new
       cxt.load(File.join(site.source,
         'assets', 'js', 'vendor', 'lunr.js', 'lunr.js'))
-      cxt.load(File.join(site.source, '_plugins', 'search.js'))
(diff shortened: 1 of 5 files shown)


_plugins/search.rb
```diff
@@ -11,14 +11,27 @@ def self.build_index(site)
         'assets', 'js', 'vendor', 'lunr.js', 'lunr.js'))
       cxt.load(File.join(site.source, '_plugins', 'search.js'))
       cxt.eval("var corpus = #{corpus_page.content};")
-      cxt.eval('corpus.entries.forEach(function(page) {index.add(page);});')
+      cxt.eval("var url_to_doc = {};")
+      cxt.eval(
+        'corpus.entries.forEach(function(page) {' +
+        '  index.add(page);' +
+        '  url_to_doc[page.url] = {url: page.url, title: page.title};' +
+        '});')
```
afeld: I'd move these lines into `_plugins/search.js`, and load that file after the corpus is passes. Less JS in Ruby files!

```ruby
cxt = V8::Context.new
cxt.load(File.join(site.source, 'assets', 'js', 'vendor', 'lunr.js', 'lunr.js'))
cxt.eval("var corpus = #{corpus_page.content};")
cxt.load(File.join(site.source, '_plugins', 'search.js'))
```

_plugins/search_hook.rb
```diff
@@ -12,8 +12,9 @@ class Site
     def after_render
       pages_api_after_render
       return if self.config['skip_index']
-      index = Hub::SearchIndexBuilder.build_index(self)
+      index, url_to_doc = Hub::SearchIndexBuilder.build_index(self)
       self.pages << index if index != nil
+      self.pages << url_to_doc if url_to_doc != nil
```
afeld: Why would the page be `nil`? I'd simplify to

```ruby
search_pages = Hub::SearchIndexBuilder.build_index(self)
self.pages.concat(search_pages)
```

_test/search_test.rb
```diff
@@ -9,5 +9,10 @@ def test_index_built
       assert(File.exist?(File.join(SiteBuilder::BUILD_DIR, 'index.json')),
         "Serialized lunr.js index doesn't exist")
     end
+
+    def test_url_to_doc_map_built
+      assert(File.exist?(File.join(SiteBuilder::BUILD_DIR, 'url_to_doc.json')),
+        "Serialized URL-to-doc info index doesn't exist")
+    end
```
afeld: We should actually check the contents, e.g. that it has `url` and `title` on every entry.

afeld:
Bummer that index.json doesn't include this information :confused: 

Forgot to check this yesterday, but `index.json` is 759k, and `pages.json` is 250k. Since the `url_to_doc.json` is 50k, this means we are now sending 559k more over the wire than we were before https://github.com/18F/hub/pull/278. Pretty sure that's still worth the tradeoff for the CPU cycles, but worth noting.

Good with it otherwise!

mbland:
Just pushed a commit with all your suggestions, @afeld.

Regarding the wire size, where are you getting the `index.json` and `pages.json` sizes from? This is what I'm seeing:

```
$ ls -l _site/index.json _site/url_to_doc.json 
-rw-r--r--  1 _ _  4116891 May  1 17:04 _site/index.json
-rw-r--r--  1 _ _    50932 May  1 17:04 _site/url_to_doc.json

$ ls -l _site/api/v1/pages.json
-rw-r--r--  1 _ _  675772 May  1 17:04 _site/api/v1/pages.json
```
So, arguably it's worse than you describe. :-)

But, good news. We have gzip compression turned on in nginx. I don't yet know how to easily see the size of the files as sent by nginx, but as an approximation:

```
$ gzip _site/index.json _site/url_to_doc.json
$ ls -l _site/index.json.gz _site/url_to_doc.json.gz 
-rw-r--r--  1 _ _  515187 May  1 17:04 _site/index.json.gz
-rw-r--r--  1 _ _    9504 May  1 17:04 _site/url_to_doc.json.gz
```

And, though we aren't doing anything to explicitly set `If-Modified-Since` headers, Chrome, at least, seems smart enough to grab the files from the cache on reload:

chrome-json-caching

So, all in all, we are pumping more data over the wire for initial requests, but compression and caching mitigates it somewhat, and we're not blocking the Javascript thread with index rebuilds after events fire. Seems like a winning trade-off. :-)

PR #281, opened 2015-05-03
Compress Javascript files

mbland:
cc: @afeld

Since asynchronous loading was introduced in #260 and #262, we lost the
Javascript compression for most scripts provided by jekyll-assets. This plugin
compresses all of the Javascript assets under `assets/js` that aren't already
compressed (i.e. end in `.min.js`).

The performance impact on `localhost:4000` may not be noticeably
earth-shattering, but appears favorable, especially in terms of earlier
DOMContentLoaded (blue line) and Window Load (red line) events.

In terms of the difference in file sizes across all scripts, there's an 82.3%
reduction:

```
$ find assets -type f -name '*.js' | grep -v '\.min\.js' | xargs ls -l | awk
'{SUM+=$5} END {print SUM}'
1029992

$ find _compressed -type f -name '*.js' | grep -v '\.min\.js' | xargs ls -l |
awk '{SUM+=$5} END {print SUM}'
182786

$ python -c 'print 182786 / 1029992.0'
0.17746351428
```

The biggest file, angular.js, is 85.6% smaller:

```
$ ls -l angular*.js
-rw-r--r--  1 _ _  113536 May  3 14:35 angular-compressed.js
-rw-r--r--  1 _ _  789267 May  3 14:35 angular.js

$ python -c 'print 113536/789267.0'
0.143849926577
```

Using command-line gzip to approximate webserver compression, compressing the
pre-compressed Javascript is still a significant win (78.8% smaller):

```
$ gzip angular*.js
$ ls -l angular*.js.gz
-rw-r--r--  1 _ _   41638 May  3 14:35 angular-compressed.js.gz
-rw-r--r--  1 _ _  196728 May  3 14:35 angular.js.gz

$ python -c 'print 41638/196728.0'
0.211652637144
```

These images show the performance of this change on top of #280, but this PR does not contain those changes:

**Uncompressed Javascript:**

uncompressed-timeline
uncompressed-network

**Compressed Javascript:**

compressed-timeline
compressed-network


afeld:
Would putting jekyll-assets back in (but not using it for concatenation or SHA URLs) achieve the same thing?

mbland:
I thought it depended on using Liquid tags to include them, but I'd love to be wrong. This is, admittedly, a bit of a hack. I'll check and see.

mbland:
OK, I figured out a way to get jekyll-assets to work again. I reverted the original commit and pushed a new one. `main.js` isn't compressible under this scheme, but it's a tiny price to pay.

mbland:
Wait...something appears funky with search...looking...

mbland:
No, it's fine, false alarm... (I'd tried a fresh repo and forgot to init the submodules.) This is good to go if you're OK with it, @afeld.

mbland:
Although, before you do that, can you maybe pull it and check out the performance numbers? I'm noticing worse performance on my machine relative to the original plugin I posted here.

mbland:
Hmm, no...sorry for the spam, but it seems the JS performance is looking equally questionable between this latest commit, `master`, and the original commit.

mbland:
Ah, I've got Chrome extensions in my one Chrome persona interfering with the times. Performance makes sense in the other Chrome persona without the extensions.

afeld:
Haha good talk :laughing:

PR #282, opened 2015-05-04
Compress all the Javascript things

mbland:
Yes, this is a bit obsessive, but now _all_ the Javascript things are compressed.

cc: @afeld


mbland:
Ping...

Issue #284, opened 2015-05-06
Change Tock Proxy URL


seanherron:
New URL should be:

`tock.18f.gov`

mbland:
Questions:
- Does this pertain to the URL that https://hub.18f.us/tock/ proxies to, or should https://hub.18f.us/tock/ be replaced by https://tock.18f.gov/?
- Does tock continue to need `google_auth_proxy` to manage MyUSA authentication?

I've managed to figure out how to configure nginx and the `google_auth_proxy` to make this happen, _except_ that the `client_id` and `client_secret` pair for `hub.18f.us` isn't transferable to `tock.18f.gov`.

Right this moment you can access both hosts: while the Hub will work, Tock will error out, even though the `google_auth_proxy` is redirecting properly. Running two instances is doable, but kludgey, since I'd have to rename one of the running services to `tock_auth_proxy` or something, and set up an additional restart trigger for that service when the list of authorized users change. (Or I could try to implement a file watcher in the `google_auth_proxy`.)

Apparently, since Tock is a Django app, it can integrate with MyUSA directly. That would mean adding that code into Tock, and managing the list of authorized users independently.

However, I think I can hack the `google_auth_proxy` to dynamically select client id/secret pairs based on the `Host` header, and continue to serve Tock in sync with the Hub.

Preference?

afeld:
I don't see any reason for Tock to continue being tied to the Hub, domain or auth-wise.

mbland:
Actually, with the help of bitly/google_auth_proxy#12, I _have_ managed to use a single `google_auth_proxy` instance to act as a single sign-on engine for multiple domains. Though, as part of the act, I've also started to 301 from `hub.18f.us` to `hub.18f.gov`.

So see for yourself: https://tock.18f.gov/tock/

I'll leave this open until the `/tock` endpoint is no longer necessary, and I've announced the `18f.us` => `18f.gov` redirect (after documenting what I did, of course).

PR #286, opened 2015-05-07
google_auth_proxy SSO and other config updates

mbland:
Inspired by bitly/google_auth_proxy#12.

@adelevie Please let me know if `SSO.md` in particular makes sense.

cc: @afeld @NoahKunin


mbland:
Forgot to mention; this also closes #284.

mbland:
Also, why not give @konklone a poke here, too. :-)

PR #290, opened 2015-05-12
Switch "Edit this page" link back to GitHub

mbland:
Closes 18F/hub-pages-private#114.

cc: @leahbannon @gbinal @wslack @afeld


wslack:
I disagree a bit with @gbinal; I find prose to be super-helpful if we can make it work. Just my 2 cents.

leahbannon:
@wslack Prose hasn't worked properly in years, and I'm not optimistic...

:+1:

PR #291, opened 2015-05-12
Stop restarting the google_auth_proxy

mbland:
Now that we're running an instance with bitly/google_auth_proxy#89
integrated, this is no longer necessary.

cc: @afeld @gboone


mbland:
Ping... One o' y'all mind merging?

PR #293, opened 2015-05-17
Prepare to extract search gem

mbland:
@afeld These are all behavior-preserving changes that set the stage to extract the Hub's search functionality into a gem, so that we can use it on 18f.gsa.gov. These are all good to go in as-is, even before gem extraction, pending your approval.

Also, if you've ideas about how to package `search.js`, `main.js`, `bower.json`, etc. into the gem to make it easy to include the client-side JS components into the repo that consumes the gem, I'm all ears. I'm currently thinking of copying these files into a data dir in the gem, and having a `jekyll_pages_api_search_init.rb` script copy them over, with README instructions on how to set them up for your specific repo.

As for the name, there is prior art: https://github.com/slashdotdash/jekyll-lunr-js-search. Might be other deployment ideas to steal from there, too.

cc: @gboone


afeld:
> ideas about how to package search.js, main.js, bower.json, etc. into the gem

Yeeeeah, this is tricky. I think I've seen plugins that provide an `include` with all of the relevant markup (Angular template, `<script>`s, etc.), but not sure of a good example off the top of my head. That would probably be a good place to start, if it's possible.

> copying these files into a data dir in the gem

The issue with copying files is that it makes upgrading a pain. Would be nice to support a few levels of usage:

* I want to include the Gem and something in my template to place the search box, and that's it.
* I want to customize the search weighting, but otherwise I don't care how it works under the hood.
* I want to be able to customize the appearance of the search box.
* I don't want to re-include Angular (or whatever other lib) because it is already being loaded on that page.

In other words, it would be nice if the users don't have to care how the plugin is implemented and what libraries it's using.

> https://github.com/slashdotdash/jekyll-lunr-js-search

I've been subscribed to that repo for a while but can't remember why I didn't use it at the time...might be worth figuring out what that gem does/doesn't do that we do/don't like and sending some PRs their way if we think we can combine efforts. https://github.com/christian-fei/Simple-Jekyll-Search is out there too.

Thoughts?

PR #294, opened 2015-05-18
Design overhaul

afomi:
screen shot 2015-05-17 at 7 59 10 pm

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -1,7 +1,26 @@
 <footer class="footer">
   <div class="container">
-    <div>
-      <img src="{{ site.baseurl }}/assets/images/logo-gsa.png" alt="GSA logo"/>
-    </div>
+    <ul>
+      <li>
+        <a href="{{ site.baseurl }}/team/">Team</a>
+      </li>
+      <li>
+        <a href="{{ site.baseurl }}/projects/">Projects</a>
+      </li>
+      <li>
+        <a href="https://pages.18f.gov/guides/">Guides</a>
+      </li>
+      <li>
+        <a href="{{ site.baseurl }}/how-to/">How to...</a>
+      </li>
+      <li>
+        <a href="{{ site.baseurl }}/n00b/">N00b</a>
+      </li>
+      <li>
+        <a href="{{ site.baseurl }}/about/">About the Hub</a>
+      </li>
+    </ul>
+
+    <img src="{{ site.baseurl }}/assets/images/logo-gsa.png" alt="GSA logo"/>
   </div>
 </footer>
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -75,7 +75,7 @@
 {{ sheet_url | split:"/" | last | asset }}
 {% endfor %}</style>
 <link href="{{ site.baseurl }}/assets/css/styles.css" rel="stylesheet">
-<link href="https://fonts.googleapis.com/css?family=Raleway:400,700%7COpen+Sans:400,700" rel="stylesheet" type="text/css">
+<link href="https://fonts.googleapis.com/css?family=Raleway:400,700%7COpen+Sans:300,400,700" rel="stylesheet" type="text/css">
 
 <!-- IE
 ================================================== -->
--- a/_includes/nav-main.html
+++ b/_includes/nav-main.html
@@ -1,18 +1,24 @@
 <nav class="nav">
   <ul>
-    <a href="{{ site.baseurl }}/">
-      <img src="{{ site.baseurl }}/assets/images/18f-logo-small.png" alt="18F logo">
-    </a>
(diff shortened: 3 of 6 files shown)


mbland:
@afomi Wow, thanks for doing this! Got a bunch of us really excited here. I'm eager to get this in, but I do have just a few notes that would be great if you could address them. (Also, remember, I'm mainly a backend guy, hence the potentially innocent/ignorant nature of some of my questions/suggestions.)

- The 18F logo doesn't appear on other Hub pages, making it difficult to navigate back to the home page. Could you somehow add it back to the top left corner?
- The text flow on the other pages ends up on the same level as the nav bar. Can you make it so it doesn't?
nav-and-text
- If you have Vagrant installed, use `vagrant up` and connect to `localhost:8080`, you can see that the authenticated user include doesn't play nice with the new nav bar. If that's not something you're able to easily fix, let me know and I'll see what I can do.
auth-include
- Though it's not documented, I've been telling folks to update `_hub.scss` instead of any other style files, including `_custom.scss`. Is there a way to move the change from there to `_hub.scss`? If not, don't worry about it.

Thanks! :-)

cc: @jenniferthibault @meiqimichelle

mbland:
@afomi BTW, I meant to say if you don't have time to do any of this, that's fine; I'll work in a branch to address these issues myself. Just wanted to give you a chance if you're able and willing.

afomi:
@mbland, thanks for the feedback! I should be able to get to this tonight. If it can't wait, by all means, cut a branch :smiley:

afomi:
Hi @mbland, I took a few steps :arrow_backward: on the nav, reverting it to where it was structurally, and moved the `_custom` styles to `_hub.scss` as suggested.

afomi:
Couple observations: 

* there is quite a bit of overlap between the top, main, and footer navigation link text - just enough to wonder if they should be identical or are slightly inconsistent
* there are couple [font-weights](https://github.com/18F/hub/blob/design-overhaul/assets/_sass/_custom.scss#L788) that don't appear to be in the project
* I wasn't sure if the site should or shouldn't be responsive.  I saw media-queries sprinkled throughout.  And I found 1 reference to a `.col`* style class. My inclination was to throw-in Boostrap's grid to handle the nav and structural elements, but I was hesitant since it is more of decision for the team.
* your responsiveness was welcoming and encouraging as a participant/contributor

:tada:

mbland:
Ah, sorry to see the nav stepped back for now, but this is a great step forwards still. To your observations:

- Yep, we should probably better organize those. In fact, "Guidelines" may not be the right header in the top nav. Open to info architecture ideas.
- Yeah, the original styles (everything not in `_hub.scss`) I copied from 18F/18f.gsa.gov when I got started. We probably could and should rethink how styles are set up here; ideally we'd like to share them between the Hub and 18f.gsa.gov, and now that we have more experience with git submodules, maybe that's the way to go.
- Ideally the site should be responsive, but we just haven't had time or resources to make it so yet. The primary use case is the desktop, by far, but it would be nice to do it right. If you want to send a PR with your Bootstrap ideas, please do!
- Glad you feel welcome! For us, it's like Santa Claus showed up way early.

Thanks so much for this! Gonna merge now. Like I said, if you've more ideas you'd like to try and send our way, you are more than welcome to do so!

meiqimichelle:
Hey all!

Thanks so much @afomi for your help! Lemme try to add to Mike's answers:

- Some work still needs to be done on the design, for sure! Thanks for the note.
- Good point on the font weights -- have to decide which ones we're using overall. The new weights reflect ongoing design changes on the 18f.gsa.gov site that we're trying to move all our guides/hubs/allTheThings towards ;) Like @mbland said, we're still trying to work out the best way to share styles across our projects.
- @mbland is right, we had limited time when we coded the site and so we only created a desktop version. The responsive code you see is left over from the style guide that we imported the basic look and feel from. However, he was a little off on the direction -- we're explicitly *not* using Bootstrap. We're using the 'Thoughtbot Suite' -- Bourbon/Neat. I'm talking with @mbland now about trying to set aside some clean-up time for the frontend code to make it easier for lovely people like yourself to contribute and not run into...."so, about that strange responsive code..." ;)
- Agree on the happy feelings! I was like :exclamation: :exclamation: :exclamation: :confetti_ball: when your help came in !!!

afomi:
hi @meiqimichelle, thanks for the info! I've only played with Bourbon and Neat a bit.

And regarding the font-weights: (a little technical note) - when playing in the dev tools, I noticed that open-sans really only reflected a few different weights (even after playing around and loading every weight 100-700). 100-300 were light. 400-500 were normal. 600-700 were bold. 

^ I thought the above was a bit odd, since I've seen more variance in the font-weights before, but just wanted to share. I didn't dig too far all the CSS to noodle it out. 

Nevertheless, 300 rendered as light, which is why I added that weight.

- Happy Tuesday

PR #295, opened 2015-05-19
Redirect from *.18f.us to *.18f.gov

mbland:
Closes 18F/tock#124.

Already running in production.

cc: @seanherron @jpyuda


afeld:
Funny timing...created this like an hour ago:

https://trello.com/c/EiZrifqo/82-have-document-a-way-to-redirect-from-old-domains

PR #303, opened 2015-06-01
submodule update instructions for #302

### LINKED ISSUE #302: Unclear how to setup git submodules
I just pulled down the repo and it seems like an important setup step is missing in the README: setting up git submodules.

Since hub relies on two submodules (and I personally have very little experience with them) it would be nice if the setup instructions included this

jeremiak:
Quick update for the README to include instructions on updating the git submodules to close #302


mbland:
Weird. That used to be in the README. Will have to figure out what happened to it.

PR #305, opened 2015-06-02
Switch Midas URL

dhcole:
Midas has a new home at https://openopps.digitalgov.gov


afeld:
Is the database the same? In other words, do we need to tell people to move their profiles over?

dhcole:
Yea, DB is the same. Everything is the same except the domain.

PR #310, opened 2015-06-05
[WIP] Collect data from doc collections, not _data

mbland:
This is a first proof-of-concept that works for team member data. In a nutshell, `_data/private/team` is now completely supplanted by data gathered from `_18F-private/team/*.md`. I haven't checked those files in, but generated them locally via:

```shell
$ mkdir -p _18F-private/team
$ cd !$
$ for f in ../../_data/private/team/*.yml
> do
> name=$(basename $f | sed -e 's/.yml$//')
> cat $f >> $name.md
> echo '---' >> $name.md
> done
```

The idea is to move away from `_data/private` and `pages/private`, and to unify "private" data and docs in one private [collection](http://jekyllrb.com/docs/collections/) repository (that can be filtered to create a "public" version). I _think_ this may make it easier for people to understand the structure of the Hub and its data, and to make updates to it. I'm hoping collection pages with front matter embedded may prove easier to work with using WYSIWYG editors as well.

I got the idea after @gboone's desire [to merge imported data into front matter for team bios in 18f.gsa.gov](https://github.com/18F/18f.gsa.gov/pull/883). Basically, this PR goes the other way, pulling data from the collections that can then be manipulated and ultimately exposed via the API.

Thoughts?

cc: @afeld @meiqimichelle


afeld:
:+1:

_plugins/joiner.rb
```diff
@@ -139,17 +139,36 @@ def self.flatten_index(index)
...
+        doc.data['title'] = doc.data[doc.data['title_field']]
+      end
+
+      if @public_mode
+        HashJoiner.remove_data data, 'private'
+      else
+        HashJoiner.promote_data data, 'private'
+      end
+
+      data.each {|k,v| @site.data[k] = v}
```
afeld: Minor: you might be able to do `@site.data.merge!(data)`.

http://ruby-doc.org/core-2.2.0/Hash.html#method-i-merge-21

mbland:
Quick update: Turning all of the `_data/private` artifacts into collections is making a lot of sense, making the code that remains a lot smaller and easier to understand. I haven't pushed all my (rebased) changes to this branch yet, but the `_plugins` code keeps getting smaller and smaller.

In the longer run, I think this will make it easy to use Jekyll to create a separate 18F team API site (the present-day `/api` endpoint family, maybe in an 18F/team-api repo?), that the Hub could then consume to generate pages (and maybe dynamically update them), per @arowla's in-person suggestion. That separate API site could import its data all kinda ways (from Midas, from `.about.yml` files), which the Hub could consume as JSON, keeping the Hub code and content itself fairly straightforward.

I've come to realize that the Hub is just a very lightweight, organized, searchable doc collection system with a (very primitive) frontend for team API data.

mbland:
Officially moved this effort to [18F/team-api](https://github.com/18F/team-api/).

PR #313, opened 2015-06-11
Update gems, but lock down redcarpet at 3.2.3

mbland:
Turns out redcarpet 3.3.1 causes intermittent null pointer segfaults. See also: vmg/redcarpet#494

Example output:

```
$ ./go serve
Configuration file: /Users/michaelbland/src/18F/hub/_config.yml
            Source: /Users/michaelbland/src/18F/hub
       Destination: /Users/michaelbland/src/18F/hub/_site
 Incremental build: disabled
      Generating...
/usr/local/var/rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/jekyll-3.0.0.pre.beta2/lib/jekyll.rb:154:
[BUG] Segmentation fault at 0x000000000e0000
ruby 2.2.0p0 (2014-12-25 revision 49005) [x86_64-darwin14]

-- Crash Report log information --------------------------------------------
[...snip...]
```

cc: @afeld @msecret @maya @gboone @dhcole

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 
 gem 'jekyll', '3.0.0.pre.beta2'
-gem 'redcarpet'
+gem 'redcarpet', '3.2.3'
 gem 'RedCloth'
 gem 'bourbon'
 gem 'jekyll-sitemap'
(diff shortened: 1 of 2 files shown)


gboone:
We definitely ran into the null pointer segfault on 18F/18f.gsa.gov earlier this week. It was a fun adventure in hunting down which Gem was causing it.

PR #323, opened 2015-06-19
Provide a better diagnosis of 403 errors

mbland:
I'll push the updated nginx config to production once this propagates to the Hub.

cc: @afeld @wslack

PR #324, opened 2015-06-20
Fix nginx config; make 403 page style standalone

mbland:
The previous changes in #323 didn't actually do the trick, as the 403 is
coming from the oauth2_proxy, not directly from nginx. This change allows the
auth.18f.gov virtual server to intercept the 403 from oauth2_proxy during
the sign-in process, after the user has successfully logged into MyUSA but
using an account that isn't authorized to access the Hub.

The `auth.conf` config uses a Server-Side Include to serve a 403 page with a
custom link. This custom link allows the user to automatically log out of
MyUSA, and will also redirect to the intended URL once the user has logged
back in using the correct account. By changing the Hub's application URL in
MyUSA to https://auth.18f.gov/, this configuration works for all 18F
properties authenticated using the Hub's oauth2_proxy (including
https://tock.18f.gov/ and https://pages-staging.18f.gov/).

Since an unauthenticated user will not be able to load scrips and stylesheets,
there is a new http_error layout that cuts most of these things and inlines
the minimum styles necessary to maintain a consistent look (error.scss).

Also adds a helpful link to check the veracity of the Shaxper quote.

@afeld @wslack This is already running in production, and I've checked it several times. Ready to be merged whenever you're able.

cc: @leahbannon, since you hit this same problem a day before Will did.

### DIFF
--- a/_layouts/http_error.html
+++ b/_layouts/http_error.html
@@ -1,0 +1,15 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+<meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1">
+<title>18F &mdash; {{ page.title }}</title>
+<style>{% asset normalize.css %}{% asset fonts.css %}{% asset error.css %}</style>
+<link href="https://fonts.googleapis.com/css?family=Raleway:400,700%7COpen+Sans:300,400,700" rel="stylesheet" type="text/css">
+</head>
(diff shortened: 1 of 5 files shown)


mbland:
Oh yeah, here's a screenshot that shows how the target URL is encoded in the page URL and in the link at the bottom of the page (I was hovering over it when I took the screenshot):

better-403

wslack:
Beautiful! Merging now.

PR #325, opened 2015-06-21
Link directly to MyUSA account swap page from 403

mbland:
In #324, the link at the bottom of the page will take the user to the oauth2_proxy's sign-in page, which will then take the user to MyUSA to swap accounts before redirecting back to the intended page. This nginx config tweak enables the link to be built such that it leads directly to MyUSA while preserving the redirect to the intended page.

Also tweaked the 403 page's language to remove direct references to MyUSA. That much less information for the user to worry about when faced with a 403 from the Prince of Denmark.

cc: @wslack @afeld @leahbannon 

Also cc: @adelevie @yozlet @harrisj just because you might get a kick out of these neat Nginx tricks in this PR and #324 
. :-)

### DIFF
--- a/deploy/etc/nginx/vhosts/auth.conf
+++ b/deploy/etc/nginx/vhosts/auth.conf
@@ -21,7 +21,14 @@ server {
   location = /403/index.html {
     ssi on;
     root /home/ubuntu/hub/_site;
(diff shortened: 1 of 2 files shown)


afeld:
I don't know anything about nginx config but

yolo

Issue #329, opened 2015-06-24
Hub edit links on private pages 404


wslack:
The edit links on the hub point to the wrong repo, so they 404. Shoutout to Becky for reminding me you can github from mobile!

mbland:
Just added a commit to #330 to fix this.

PR #330, opened 2015-06-24
Complete _pages/private submodule migration

mbland:
Should've been part of #327. Updates to https://hub.18f.gov/private aren't appearing because the repo on that server is pointing at an outdated submodule hash.

cc: @afeld @CaliforniaKara


mbland:
Just pushed another commit to close #329 while we're at it.

PR #332, opened 2015-06-25
Adapt Jekyll::StaticFile for collections, defaults

mbland:
This enables files such as images and PDFs to show up in the same relative
output directory as other HTML and Markdown documents in the same collection.

It also enables static files to be hidden using defaults from _config.yml in
the same way that other documents in the same collection and directories may
be hidden using `published: false`, enabling the removal of
`Hub::Joiner::filter_private_pages`.

cc: @afeld @arowla

( I also wonder if it might make sense to contribute this upstream.)


mbland:
Also cc: @gboone.

Hmm, the test failure is because I forgot Travis can't access `_pages/private`. Fixing...

mbland:
Fixed.

mbland:
Looks like jekyll/jekyll#3823 is set to go in soon, but I'm not sure when. In the meanwhile, anyone care to do the honors of merging this? I'll rip out the monkey patch in a later PR once the updates to `Jekyll::StaticFile` appear upstream.

gboone:
I say this looks good. Merging!

mbland:
OK, so jekyll/jekyll#3823 is now in! Figure I'll wait for a new Jekyll gem to get cut that includes the change, then I'll undo the `Jekyll::StaticFile` monkey patch from this PR.

afeld:
Maybe just add an inline comment saying `remove this when https://github.com/jekyll/jekyll/pull/3823 is released`?

PR #335, opened 2015-06-29
Fix /skills and /interests pages

mbland:
At some point, and I'm not sure if it was a Jekyll upgrade or the move to turn
_pages into a collection, the scripts specified in _layouts/skills.html were
no longer inherited by _pages/skills.html and _pages/interests.html, causing
both pages and pages generated dynamically from them to be blank. Listing
the scripts directly in the page objects resolves the issue.

Took the liberty to reformat _includes/scripts.html as well.

cc: @afeld @dhcole @gboone @jeremiak

### DIFF
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -41,10 +41,8 @@
 <!-- Browser warning, console message -->
 <script src="{% asset_path 18f.js %}"></script>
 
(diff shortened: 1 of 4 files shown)


mbland:
Ping... will merge myself shortly if everyone else is swamped.

PR #338, opened 2015-06-30
Enable anyone at gsa.gov to view the 18F Hub

mbland:
Adds a default auth include, used by nginx's `try_files` directive when a pregenerated auth include file for the authenticated user doesn't exist. Updates the existing auth include layouts slightly.

Already successfully tried it out on https://hub.18f.gov/ by manually updating the site and removing my own specific entries in `/auth/`.

cc: @gbinal @afeld @gboone

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -38,6 +38,7 @@ exclude:
 - Vagrantfile
 - bower.json
 - bower_components
+- coverage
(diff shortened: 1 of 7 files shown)


gbinal:
:+1:

PR #345, opened 2015-07-09
Eliminate old components, update gems

mbland:
Per: https://gemnasium.com/18F/hub

All of the old Bower stuff is now encapsulated in the [jekyll_pages_api_search Gem](https://github.com/18F/jekyll_pages_api_search/).

cc: @afeld @monfresh


monfresh:
You still have an open security alert for `redcarpet`. You have it locked at `3.2.3` but it needs to be updated to `3.3.2`. You probably just ran `bundle update` and called it a day :smile:

mbland:
What, that didn't work? Isn't it supposed to?

PR #346, opened 2015-07-09
Update Jekyll and Redcarpet gems

mbland:
Meant to update Redcarpet in #345, but failed to notice the version was
locked.

cc: @afeld @monfresh


monfresh:
Sweet! All green (except for jekyll-assets, which can't be upgraded because it only works with jekyll 2.x).

PR #347, opened 2015-07-13
Change Greg Godbout to Greg Boone

awfrancisco:


afeld:
@gboone Why didn't you *tell* us you were formerly Greg Godbout???

gboone:
Shhhh! There's a reason we were never in the same room.

wslack:
This is my favorite PR title. :D

mbland:
FYI, we may be down to one Greg, but we've now got two Diegos.

PR #348, opened 2015-07-15
Update _typography.scss and _buttons.scss

andrewmaier:
Only apply antialiasing to high-resolution displays.

### DIFF
--- a/assets/_sass/base/_typography.scss
+++ b/assets/_sass/base/_typography.scss
@@ -1,5 +1,14 @@
+@media 
+(-webkit-min-device-pixel-ratio: 2), 
+(-webkit-min-device-pixel-ratio: 192), 
+(min-resolution: 192dppx) {
+  body{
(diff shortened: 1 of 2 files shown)


andrewmaier:
@mbland

mbland:
So it turns out there's two hits in `assets/css/fonts.css` as well:

```shell
$ find assets -type f | xargs grep -i antialias
assets/_sass/base/_typography.scss:  -webkit-font-smoothing: antialiased;
assets/_sass/base/extends/_button.scss:  -webkit-font-smoothing: antialiased;
assets/css/fonts.css:  -webkit-font-smoothing: antialiased;
assets/css/fonts.css:  -webkit-font-smoothing: antialiased;
```

Worth updating those, too?

andrewmaier:
@mbland Looks like they're just icons, which I'm okay with until we move to SVG :)

PR #354, opened 2015-07-21
Import updated Chart.min.js; tweak financials.js

mbland:
So the original copy of Chart.min.js was causing a bizarre error that's not
worth going into (pull this commit's parent if you want to see it). Basically,
it _insists_ on loading from a parent directory called `Chart.js`. Really.
Search for `p.push.apply` in Chart.js, ye mighty, and despair.

Also took out the `console.log()` statements from financials.js, and replaced
the chart name calculation with a plain array. Will follow-up with a commit to
financials.md in 18F/hub-pages-private.

cc: @arowla @jmcarp @monfresh


mbland:
Oh, and proof:

unminified-chart-js

assets/js/financials.js
```diff
@@ -28,22 +28,15 @@ var munge_data = function(chart_data, chart_name) {
   return data_out;
 };
 
-var raw_data = $('#chart-data').data('chart');
-console.log(raw_data);
-var chart_names = Object.keys(raw_data[0]).filter(function(x) { return x != 'name' });
+var chart_names = ['revenue', 'expenses', 'billable_hours'];
 var charts = [];
 chart_names.map(function(chart_name) {
```
jmcarp: Minor, but it's more idiomatic for the function passed to `map` to return values than to mutate some external data structure, like this:
```javascript
var charts = chart_names.map(function(chart_name) {
  var ctx = $('#' + chart_name + '-chart').get(0).getContext('2d');
  var munged_data = munge_data(financial_chart_data, chart_name);
  return Chart(ctx).Bar(munged_data, {
    'scaleShowHorizontalLines': true,
    'scaleShowVerticalLines': true,
    'scaleBeginAtZero': true
  });
});
```

mbland:
Looks like this is good-to-go; merging.

PR #355, opened 2015-07-23
Generalized chart munging JS into one file

arowla:
- Takes a chart_set config var initialized in the Markdown for the page


assets/js/chart_munger.js
```diff
@@ -0,0 +1,44 @@
+var munge_data = function(chart_data, chart_name) {
+  // var config = chart_data.config;
```
mbland: Yank out the vestigial comments?

assets/js/chart_munger.js
```diff
@@ -0,0 +1,44 @@
...
+    var label_lower = label.toLowerCase();
+    label_lower = label_lower.replace(' ', '_');
+
+    data.label = label;
+
+    // $.extend(data, config[label_lower]);
+
+    data.data = data_in.map(function(item) {
+      var value = chart_set.data_item_func(item, chart_name, label_lower);
+      return parseInt(value) % 1 === 0 ? parseInt(value) : 0;
```
mbland: Curious why this is necessary. Maybe add a comment? Also, the premature optimizer in me wants to set `value = parseInt(value);` right before this line.

assets/js/chart_munger.js
arowla: Well I've been assuming all along that we'd want to go back in and add color specifications per-dataset (like to distinguish "Actual" and "Projected" by color on the bar chart), so I left those in. I had it working pre-refactor, but hadn't gotten around to reimplementing it. I can take them out.

assets/js/chart_munger.js
```diff
@@ -0,0 +1,44 @@
...
+    var data = {};
+    var label_lower = label.toLowerCase();
+    label_lower = label_lower.replace(' ', '_');
+
+    data.label = label;
+
+    // $.extend(data, config[label_lower]);
+
+    data.data = data_in.map(function(item) {
+      var value = chart_set.data_item_func(item, chart_name, label_lower);
```
mbland: So `chart_set` is effectively global because it's in `window`?

assets/js/chart_munger.js
arowla: This was to set null (which after `parseInt` becomes NaN) to 0. But come to think of it I wrote that before the charts were working in full swing, so it might be worth checking whether this is necessary.

assets/js/chart_munger.js
mbland: Maybe replace it/preface it with a `TODO(arowla):`?

arowla:
@mbland want to take a final look? Should be all fixed up.

PR #356, opened 2015-07-28
Multiple chart config support for chart munger

arowla:
Also refactored for clarity, and to support excluding outlier data points.

Co-dependent with 18F/hub-pages-private#166


mbland:
Ah, we really need to start adding JavaScript tests soon... :-)

arowla:
Yesssss.... Was kicking myself for not having any, but I've never done client-side Javascript testing and didn't want to throw that monkey wrench into the mix just yet...

PR #358, opened 2015-07-30
Proxy to HTTPS

jmcarp:
Since we're now requiring SSL on all Cloud Foundry apps, the `proxy_pass` to http://tock.cf.18f.us is failing. This patch proxies instead to https://tock-app.18f.gov. Note: I added the "app" to distinguish between the nginx URL (tock.18f.gov) and the backend URL (tock-app.18f.gov).

ping @mbland @ramirezg @dlapiduz

### DIFF
--- a/deploy/etc/nginx/vhosts/tock.conf
+++ b/deploy/etc/nginx/vhosts/tock.conf
@@ -24,7 +24,7 @@ server {
   port_in_redirect off;
 
   location / {
-    proxy_pass http://tock.cf.18f.us;
+    proxy_pass https://tock-app.18f.gov;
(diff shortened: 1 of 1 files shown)


dlapiduz:
:+1: please merge asap

mbland:
FYI, these configs aren't automatically deployed. But I've gone ahead and deployed this config now.

dlapiduz:
Cool! Tock is back up! Thanks all

mbland:
Thanks for fixing the "bug"! Was making me nervous, proxying over http, even if it technically is in the same cluster...

PR #359, opened 2015-08-03
Fix Tock proxy issue discussed in 18F/tock#219

mbland:
Already running in production.

cc: @jmcarp @ramirezg @batemapf @ethanheppner @konklone @melodykramer @afeld @NoahKunin

### DIFF
--- a/deploy/etc/nginx/vhosts/tock.conf
+++ b/deploy/etc/nginx/vhosts/tock.conf
@@ -24,7 +24,9 @@ server {
   port_in_redirect off;
 
   location / {
-    proxy_pass https://tock-app.18f.gov;
+    proxy_pass https://tock-app.18f.gov/;
+    proxy_http_version 1.1;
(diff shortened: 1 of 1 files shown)


mbland:
As for why this didn't show up before the switch to https://tock-app.18f.gov, my theory is that the old setup was only sending relative URLs in the 302, where the new setup is including the protocol and host. In other words, I think it went from this:

```
HTTP/1.1 302 Found
Location: /
```

to this:

```
HTTP/1.1 302 Found
Location: https://tock-app.18f.gov/
```

The nginx logs on the hub don't provide any insight; they show a couple of 500s, but mostly 302s after a `POST /reporting_period/2015-07-26/ HTTP/1.1` operation, but not what that 302 was redirecting to. Would be nice to dig in to see if this theory is accurate, and I'm happy to do so if people wish, but for now I'm going to move along.

konklone:
:+1:

PR #361, opened 2015-08-03
Javascript Unit Tests for Chart Munger

arowla:
Tests for the chart munger functionality. Munger has also been refactored to _be testable_. To run, load up the hub at `/tests/`.

Mocha and Chai have been vendorized in `assets/js/tests`, and they can probably stay that way, but I'll soon take a crack at adding a postInstall step to package.json to copy them there automatically so they can be kept up to date with an `npm install`.


arowla:
Dependent on 18F/hub-pages-private#184

mbland:
Neat!

PR #363, opened 2015-08-05
Update the blogging page

gboone:
Clarifies _where_ the issue should be opened, and provides a handy link to opening a new issue. It'd still be cool if we could auto-populated the "new issue" form. Some day…

Thanks @zacharycohn for bringing this poor UX to our attention. Hopefully this will make the process easier for folks in the future.


gboone:
cc: @18F/blog

PR #368, opened 2015-08-07
Added Javascript vendorizing go script command

arowla:
I did not add anything to handle the case when node/npm is not installed. Let me know if that seems important.

### DIFF
--- a/go
+++ b/go
@@ -63,6 +63,12 @@ def update_gems
   exec_cmd 'git add Gemfile.lock'
 end
 
+def update_js
+  exec_cmd 'npm update'
+  exec_cmd 'npm install'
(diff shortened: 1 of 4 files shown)


go
```diff
@@ -63,6 +63,12 @@ def update_gems
   exec_cmd 'git add Gemfile.lock'
 end
 
+def update_js
```
mbland: Maybe put something like this as a sanity check:

```ruby
  abort 'Install npm to update JavaScript components: ' \
    'http://nodejs.org/download/' unless system 'which npm > /dev/null'
```

arowla: Done, and confirmed that it works.

PR #381, opened 2015-08-21
Use go_script gem to refactor ./go script

mbland:
cc: @afeld @gboone @arowla


mbland:
Ping... Updated this to use go_script v0.1.0.

PR #385, opened 2015-08-26
Update Hub tutorial permalink

mbland:
Also tweaked the front matter example.

cc: @afeld @wslack

### DIFF
--- a/_pages/qa/hub-tutorial.md
+++ b/_pages/qa/hub-tutorial.md
@@ -1,6 +1,6 @@
 ---
 layout: q-and-a
-permalink: /tutorial-for-prose-and-hub/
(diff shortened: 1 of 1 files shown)


wslack:
If the layout isn't set in the frontmatter, where is it set?

afeld:
@wslack What do you mean? It's set at the top of the file, though there's also [a default](https://github.com/18F/hub/blob/cd675242d787fd0cd6cf71f05b29b680d779a412/_config.yml#L88).

wslack:
I was being foolish; I didn't realize the tweaked example was "smart" like that and thought we were asking people to copy that. Thanks @afeld!

PR #387, opened 2015-08-26
fix broken links

afeld:
See commits.


mbland:
Looks like Travis is bummed...

afeld:
I think the tests are failing on Travis but not locally (on Mac) because of the case-sensitive filesystem :rage: Will investigate in a few hours, unless you have a moment to look at it sooner?

mbland:
Wish I did, but if you don't get to it by the next time I look, I'll look as well. (Wonder if Travis has a case-insensitive FS setting?)

afeld:
Phew, ok, fixed by canonicalizing (downcasing) the user `name`s across the various templates. :boom:

PR #388, opened 2015-08-26
remove old data when importing

afeld:
Broken out from #387.


mbland:
Whoops, missed this yesterday...

PR #394, opened 2015-08-27
Removes emoji support

gboone:
Cuz, well, it was never really supported. I think this is all we'll need.

PR #396, opened 2015-08-27
Removes all the bower components I accidentally committed

gboone:
What it says on the can.


mbland:
Appreciate it, but I already pushed a commit to do this onto #394.

PR #399, opened 2015-08-31
Update n00b.md

andrewmaier:
Change the "help make onboarding better" copy to not assume GitHub knowledge and to also reflect that @andrewmaier and @melodykramer are point people on this.


andrewmaier:
@mbland 👆

PR #401, opened 2015-09-02
Attempt to fix styling, especially h1-h4

mbland:
I've tried to take a pass at tweaking some of the styles for headlines and unordered lists to make pages like https://hub.18f.gov/consulting/interviewing/core-values/ more appealing.

Before:

before

After:

after

However, my CSS/design capabilities are _extremely_ limited, and would love it if a true expert like  @noahmanger,  @meiqimichelle, or @jenniferthibault might have a couple spare cycles to do this up properly.

Or, @adelevie, if you think this gets us in the ballpark of what you and yours had in mind, feel free to merge.

### DIFF
--- a/assets/_sass/_hub.scss
+++ b/assets/_sass/_hub.scss
@@ -16,6 +16,7 @@
 //********************************************************
 ul {
   list-style: disc inside none;
+  padding-bottom: 0.5em;
 
   li {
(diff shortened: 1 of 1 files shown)


adelevie:
Hey @mbland I think this looks great. Thanks for doing this!

PR #404, opened 2015-09-03
Improves training instructions

emileighoutlaw:

PR #405, opened 2015-09-03
fixes training permalink

emileighoutlaw:


mbland:
Actually @emileighoutlaw, can you do this in #404? You can edit the file directly via the GitHub interface to push a new commit.

PR #410, opened 2015-09-07
Remove line on signal-noise in USDS Slack section

konklone:
This is superfluous for this section of the Slack standards - if it's only about signal-noise ratio, that applies to basically all of our on-topic channels, no matter who's in them.


konklone:
cc @NoahKunin

PR #412, opened 2015-09-08
Added a link to Tock FAQ

StephaniePlageman:


wslack:
Hey @afeld, what is policy with Google Doc links like this on a public page? (if this is a public page?)

afeld:
It's not a _huge_ deal since the access is controlled separately, but we should [figure out a better way to handle them](https://github.com/18F/hub/issues/413).

afeld:
p.s. thanks @StephaniePlageman!

mbland:
Sorry I didn't catch this sooner, but ["click here"-style links aren't very accessible](https://pages.18f.gov/accessibility/links/). We also [prefer active voice to passive](https://pages.18f.gov/content-guide/active-voice/).

Since I'm on vacation, I'll resist the urge to update this myself, but something like this may be better:

```
Learn more about Tock from the [Tock FAQ](...).
```

wslack:
@StephaniePlageman

PR #441, opened 2015-09-18
add dolores.conf

gemfarmer:
As per your comment on the dolores-bot repot (https://github.com/18F/dolores-landingham-bot/issues/2#issuecomment-141159857), I made an nginx .conf file. Am I missing anything?

### DIFF
--- a/deploy/etc/nginx/vhosts/dolores.conf
+++ b/deploy/etc/nginx/vhosts/dolores.conf
@@ -31,31 +31,3 @@ server {
     proxy_set_header Host $host;
   }
 }
-
-# Staging server
-server {
-  listen 80;
-  server_name  dolores-staging.18f.gov;
-  return 301 https://$host$request_uri;
-}
-
-server {
-  listen 443 ssl spdy;
-  server_name  dolores-staging.18f.gov;
-  include ssl/star.18f.gov.conf;
-  include vhosts/auth-locations.conf;
-}
-
-server {
-  listen 127.0.0.1:8080;
-  server_name dolores-staging.18f.gov;
-  port_in_redirect off;
-
-  location / {
-    resolver 8.8.8.8;
-    set $backend_upstream "https://dolores-app-staging.18f.gov";
-    proxy_pass $backend_upstream;
-    proxy_http_version 1.1;
-    proxy_set_header Host $host;
-  }
-}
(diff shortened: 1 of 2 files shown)


mbland:
So there is going to be a staging instance at `dolores-staging.18f.gov`?

Also need to add a line at the bottom of `deploy/etc/nginx/nginx.conf` to include the new config.

jessieay:
I don't think we need staging for this, but don't feel strongly either way.

gemfarmer:
@mbland I added the reference to `dolores.conf` in `nginx.conf` 
I was assuming that we would have a staging site, but if there is a reason why culling that would be better/easier, then we could just have a production site.

deploy/etc/nginx/vhosts/dolores.conf
```diff
@@ -0,0 +1,61 @@
...
+}
+
+server {
+  listen 127.0.0.1:8080;
+  server_name dolores.18f.gov;
+  port_in_redirect off;
+
+  location / {
+    resolver 8.8.8.8;
+    set $backend_upstream "https://dolores.18f.gov";
```
mbland: This should be the Cloud Foundry URL where the app will actually be deployed. (As it stands, it's going to try to proxy to itself.)

deploy/etc/nginx/vhosts/dolores.conf
```diff
@@ -0,0 +1,61 @@
...
+}
+
+server {
+  listen 127.0.0.1:8080;
+  server_name dolores-staging.18f.gov;
+  port_in_redirect off;
+
+  location / {
+    resolver 8.8.8.8;
+    set $backend_upstream "https://dolores-staging.18f.gov";
```
mbland: Same here re: CF URL.

mbland:
@gemfarmer Whether or not you want a separate staging instance is entirely up to you. Can always add one later if you want, but it's probably best to take it out of the config for now if you don't.

I'll set up the DNS mapping for dolores.18f.gov.

mbland:
OK, the DNS mapping is live.

deploy/etc/nginx/vhosts/dolores.conf
gemfarmer: Ok. I'm working on creating the URL. I've been running into some CF issues.

mbland:
This is now running in prod, and all is well... Except there needs to be a Cloud Foundry route from `dolores.18f.gov` to `dolores-app.18f.gov`. Sorry for not recalling the lessons from #393.

PR #442, opened 2015-09-22
link from the About page to the Editing Tutorial

afeld:
Boo duplication!

/cc @mbland @wslack @jessieay


wslack:
"Some checks haven’t completed yet" but otherwise LGTM

PR #447, opened 2015-09-23
remove dolores.conf

gemfarmer:
Hey @mbland. I have the app-level authentication set up, but it is still proxying us from `dolores.18f.gov` to `dolores-app.18f.gov`. Along with this PR to remove `dolores.conf, what needs to be done to prevent proxying from one to the other?

When that is in place I can remove the CF route to `dolores-app`, and the site can exist at dolores.18f.gov


mbland:
I just deleted the record from Amazon Route 53. Should be good to go.

gemfarmer:
Delayed thanks! @mbland

PR #450, opened 2015-10-05
clarify that PIFs should be added to Slack

afeld:
Per https://18f.slack.com/archives/admins-slack/p1443648124000836.

/cc @quepol

### DIFF
--- a/_pages/standards/Slack.md
+++ b/_pages/standards/Slack.md
@@ -28,7 +28,7 @@ If you see anyone violating our Code of Conduct, please report it immediately in
 
 ### Who can use Slack?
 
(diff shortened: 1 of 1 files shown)


wslack:
Hmmm. do we want to say its required for PIFs @afeld or that they have access as full members as needed?

afeld:
That seems like a @FreeMason question – no strong feelings from my end.

afeld:
...but I would say that we should default to including them.

afeld:
Ok, got a :+1: from Puja and @rossdakin. Merging!

stroupaloop:
Confirming on here, all PIF should be added to 18F Slack by default. Thanks!

PR #455, opened 2015-10-21
Update macbook-security.md

rossdakin:
Include link to NSA PDF


mtorres253:
Looking at this with @mbland. We're not sure what this is about. Would you add more details when you have a moment?

rossdakin:
WYSIWYG

Just something that I stumbled upon and thought may be relevant to someone reading this article who really wants to lock down their Mac.

No strong feelings about including it other than that.

nicoleslaw:
@rossdakin We've moved the main content about Mac security to the handbook. Want to add it there?: https://github.com/18F/handbook/blob/staging/articles/4-how-we-work/equipment.md#laptop

PR #480, opened 2015-11-12
Delete gsa-dell.md

andrewmaier:
This self-service documentation feels superfluous given that we don't issue Dells to new hires.


andrewmaier:
cc @mbland @NoahKunin @afeld

afeld:
> we don't issue Dells to new hires

Is that true? I'm pretty sure I got one, and @lsgitter locked it in a drawer.

andrewmaier:
Still not sure you need this content @afeld :)

afeld:
Agreed :+1:

PR #500, opened 2015-12-24
Updates to nginx configs after http/2 upgrade

mbland:
Also includes some configs not before checked in.


mbland:
Also pushed a commit to fix the broken link introduced in 3cf6563, whereby `handbook.18f.gov` without the preceding `https://` was interpreted as an internal link. cc: @melodykramer

PR #502, opened 2015-12-28
Test about yml connects 18F/about_yml#37

ertzeid:
Allow use of about_yml's `run_about_yml_check` task within Hub branches.

Also fixes the `.about.yml` file so that the above check passes.

### DIFF
--- a/.about.yml
+++ b/.about.yml
@@ -24,3 +24,11 @@ links:
 - url: https://18f.gsa.gov/hub/
   text: 18F Public Hub production instance
 tasks: https://github.com/18F/hub/issues
+team:
+- github: mbland
+- github: afeld
+- github: melodykramer
+licenses:
+  .about.yml:
+    name: CC0
+    url: https://github.com/18F/hub/blob/master/LICENSE.md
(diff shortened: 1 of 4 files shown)


mbland:
OK, so the "push" build failed because that branch is based on 06607e92375a76b5a0c7ec629d39315a5cdf6fac. That commit came before 71d4130d13fa6243a012d800cc38f2863d1b808e from #500, which fixed a breakage from 3cf65637ed283369da64249e7b6b94a6499f3dc4 whereby `handbook.18f.gov` looked like an internal URL without the `https://` prefix. You'll see the head commit of this failing build is 11527e7f9dc63bec9e373ef47860d46c8a991fed from this PR, `Enable use of run_about_yml_check rake task`.

The "pr" build passes because it represents a merge commit into `master`, which of course has the fix. The head of this build is the pending 82c0f62130514513426271c5bbbf810f6f56a8b9, `Merge 11527e7f9dc63bec9e373ef47860d46c8a991fed into 4bfe665ca903eef6d760e78f77088c301b23fa38`.

File this "failing push vs. passing PR"  knowledge under "good to know".

PR #503, opened 2015-12-29
Add .about.yml check rake task to Travis config. Connects 18F/about_yml#37

ertzeid:


.travis.yml
```diff
@@ -14,3 +14,7 @@ env:
 script: bundle exec ./go ci_build
 notifications:
   webhooks: http://project-monitor.cf.18f.us/projects/hub/status
+
+script: bundle exec rake run_about_yml_check
```
mbland: I think this should get added to the `./go ci_build` command instead.

ertzeid: I'm happy to do that, but was hoping to have an example for other teams who
wanted to automatically run this check on their own files. Is there a way
to preserve this diff in GitHub so that I can link it from the about_yml
README?

mbland: I'd think the 18F/about_yml#71 is sufficient without needing to link to a commit, but you can just keep this original commit in the branch history and it'll stay alive.

ertzeid: I'm pretty indifferent. Happy to edit that PR to remove the link, or leave
it in and leave this branch alive.

mbland: Your call! :smiley:

ertzeid: I'm gonna change 18F/about_yml#71... Update incoming.

PR #504, opened 2015-12-30
add statistics to Locations page

afeld:
Well, statistic :wink: Inspired by @hollyallen's [post](https://18f.slack.com/archives/dev/p1451433878001907) in Slack. We could probably do more clever things here, but this is a start:

screen shot 2015-12-30 at 12 47 22 am


monfresh:
LGTM

mbland:
We should consider integrating this into [18F/team_api](https://github.com/18F/team_api/) at some point.

Issue #499, opened 2015-12-23
Integrate new data-private submodule work with Hub


mtorres253:
This story is to hook up Team API to the Hub before the migration is done. The Hub will need to expect the change and get its data from the new location.

Everything that the Hub is currently pulling directly from the team and project directories are now being pulled in via a Team API response.

To Demo: Make sure the Hub looks the same or better after this work is done. It should at least contain all the data it currently contains.

To Demo: 

After data-private migration:
https://github.com/18F/team-api.18f.gov/issues/166
https://github.com/18F/team-api.18f.gov/issues/167

This should still work

mtorres253:
Hi Catherine, is this in progress?

catherinedevlin:
It is in progress!  I haven't come to a sane point to push a branch up to github yet - probably tomorrow, though.

mtorres253:
Thanks @catherinedevlin moving to in progress column

mtorres253:
Does someone need to review this or can I close it?

PR #524, opened 2016-01-26
DO NOT MERGE: Get team data from team-api server

catherinedevlin:
DON'T BOTHER.  The part of the Hub affected by this code is now being split out into https://github.com/18f/team-browser instead.

fixes https://github.com/18F/hub/issues/499

When I started replacing stuff in the hub's `_data` directory with
results taken from https://team-api.18f.gov/api/, I promised there
would be no change to the results produced.  Well...

I've been manually comparing the results of the two approaches
(it turns out that there are too many irrelevant differences to
make automated comparison possible - image files are renamed,
API results are slightly fresher than the hub).  With much manual
comparison, I assert that there are no *substantive* differences,
except... 

There's still one "irreconcilable difference": `_data` currently
includes pages for some projects that incorporate multiple github
repos, like https://hub.18f.gov/projects/openfec/.  The team_api
server assumes a 1:1 relationship between projects and github 
repos, and the project description, etc. are attached to each
github-based project.  There is an occasionally-used "parent"
field, but the parent projects that are named usually don't
even exist in the team_api data.


.ruby-version
```diff
@@ -1 +1 @@
-2.2.3
+2.2.0-preview1
```
ccostino: D'oh!  Why did this get kicked down to an older release of Ruby?  I don't believe we want that.

Gemfile
```diff
@@ -13,12 +13,14 @@ gem 'uglifier'
...
 gem 'go_script'
 gem 'about_yml'
+gem 'hmac_authentication'
 
 group :jekyll_plugins do
   gem 'jekyll-assets'
   gem 'jekyll_pages_api'
   gem 'liquid_pluralize'
   gem 'jekyll_pages_api_search'
+  # gem 'jekyll-get', :git => 'https://github.com/18F/jekyll-get.git'
```
ccostino: Should this be uncommented?  Or is it no longer needed?

_includes/team_members.html
```diff
@@ -1,14 +1,24 @@
...
       <a href="{{ site.baseurl }}/team/{{ member.name | canonicalize }}">{{ member.full_name }}</a>{% if member.location %} -
       <a class="location" href="{{ site.baseurl }}/locations#{{ member.location }}">{{ member.location }}</a><br/>{% endif %}
       {% if member.projects %}
-        Projects: {% for p in member.projects %}<a href="{{ site.baseurl }}/projects/{{ p.name | canonicalize }}">{{ p.project }}</a>{% unless forloop.last %}, {% endunless %}{% endfor %}<br/>
+        Projects:
+        {% for p in member.projects %}
+          <a href="{{ site.baseurl }}/projects/{{ p.name | canonicalize }}">
+            {{ p.name }}</a>
+            {% unless forloop.last %}, {% endunless %}
+        {% endfor %}<br/>
```
ccostino: Much cleaner and easier to read, thank you!

_layouts/skill.html
```diff
@@ -0,0 +1,22 @@
...
+	<section class="hub-page-section">
+
+	<ul>
+		{% for i in skill.members %}
+			<li><a href="{{ site.baseurl }}/team/{{ i.name | canonicalize }}">{{ i.full_name }}</a></li>
+		{% endfor %}
+	</ul>
+
+	</section>
+	{% endif %}
```
ccostino: Should we have an `{% else %}` portion to handle displaying a message that no matches were found instead of what will appear to be a blank page under the heading?

_pages/docs.md
```diff
@@ -14,7 +14,7 @@ Documentation integrated into the Hub and linked from this index should be store
 Visit our [News Clips page]({{ site.baseurl }}/clips) to see press coverage of our team, our products, and our mission.
 
 ## DigitalGov Webinar Videos
-- [Large Scale Development Culture Change: Google and the US Government](https://www.youtube.com/watch?v=CWSLSHljQLM) _by [Mike Bland]({{ site.baseurl }}/team/mbland)_
+- [Large Scale Development Culture Change: Google and the US Government](https://www.youtube.com/watch?v=CWSLSHljQLM) _by [Mike Bland]({{ site.baseurl }}/team/michael.bland)_
```
ccostino: Where is "[Pain is Over, If You Want It](https://www.youtube.com/watch?v=ahtihwxgriA)"?  Should that be included here as well?

_plugins/hub.rb
```diff
@@ -29,23 +29,34 @@ module Hub
...
+      unnest_site_api_data(site)
+      site.data['guest_users'] = []
+
       Joiner.join_data(site)
       Snippets.publish(site)
-      CrossReferencer.build_xrefs(site.data)
-      Canonicalizer.canonicalize_data(site.data)
       Auth.generate_artifacts(site)
-      ::HashJoiner.prune_empty_properties(site.data)
+      #::HashJoiner.prune_empty_properties(site.data)
```
ccostino: Should we think about trying to remove all of this?  I feel like I'm looking at the [Team API](https://github.com/18F/team_api gem) here and it seems odd that the work would have to be repeated, especially as we're pulling from the API itself now.

Perhaps this is something to happen in the future or maybe there is still a need I'm not aware of, but I wanted to at least raise the question.

_plugins/jekyll_get.rb
```diff
@@ -0,0 +1,74 @@
+require 'json'
+require 'hash-joiner'
+require 'open-uri'
+require 'hmac_authentication'
+
+
+module Jekyll_Get
```
ccostino: Oh, this was pulled in directly?  Should we just get rid of the line in the `Gemfile` then that I noted above?

_plugins/joiner.rb
```diff
@@ -127,27 +119,13 @@ def initialize(site)
...
-    end
-
-    # Joins public and private project data.
-    def join_project_data
-      promote_private_data 'projects'
-
-      if @public_mode
-        @data['projects'].delete_if {|p| p['status'] == 'Hold'}
+    # Populates details of team members in location data
+    def extend_location_data
```
ccostino: @mbland:  If we add full location cross referencing support to the `team_api` gem, would that obviate the need for this method here?

ccostino:
Thanks for working on this, @catherinedevlin!  I made several notes and had some questions, and I think @mbland should take a look as well.

One thing that did come up was that when I pulled this branch and tried to run the tests, they fail with this error:

```
hub/_plugins/api.rb:166:in `create_filtered_hash': undefined method `each' for nil:NilClass (NoMethodError)
```

ccostino:
@catherinedevlin pointed out that I needed to set the `HMAC_KEY` environment variable; that did the trick!  There is still a failing test locally, but it is failing in master as well for me:

`Hub::PagesApiTest#test_all_posts_have_titles [./hub/_test/pages_api_test.rb:61]:`

ccostino:
@catherinedevlin:  One other thing that should probably change as a part of this PR is https://github.com/18F/hub/blob/master/_data/README.md - right now it has a reference to the `data-private` repo, but this PR is removing that connection.

This would tie into the work with https://github.com/18F/team-api.18f.gov/issues/169

ccostino:
Same with https://github.com/18F/hub/blob/master/_pages/101/internal-vs-public.md, to reflect the work done in here, unless @andrewmaier and @nicoleslaw want to consider moving this to the Handbook as well.

catherinedevlin:
Made irrelevant by new team-browser repo.

PR #547, opened 2016-02-10
Set default args to empty list for `./go build`

mbland:
The https://hub.18f.gov/hub/ build was breaking because the `args` parameter was passed as `nil` instead of `[]`. Setting the default value for `args` explicitly rectifies this.

cc: @afeld @ccostino @ertzeid @catherinedevlin

### DIFF
--- a/go
+++ b/go
@@ -97,7 +97,7 @@ def_command :validate, 'Build, then validate no internal links are broken' do
 end
 
 def_command(
-  :build, 'Builds the internal and external versions of the Hub') do |args|
+  :build, 'Builds the internal and external versions of the Hub') do |args = []|
   puts 'Building internal version...'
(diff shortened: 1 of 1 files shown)


afeld:
Ah, so Travis didn't catch the problem since it only affected the internal build? Is there any test we could add that would show this fixes the problem? Since this fixes an "immediate" issue, totally fine if you want to look into that after merging this.

mbland:
Yeah, been thinking about this. The problem seems to lie with the fact that the `./go` script is what runs the git sync, so updates to the `./go` script only take effect on the build _following_ the update.

afeld:
Ok, not sure I totally understand, but let's keep brainstorming about how to catch these problems in tests.

PR #566, opened 2016-02-24
Remove ruby version

mbland:
Companion to 18F/team-api.18f.gov#214. `.ruby-version` proves too restrictive when running in a Docker container (from 18F/knowledge-sharing-toolkit) that has a different version installed.

cc: @ertzeid @ccostino @jbarnicle


ccostino:
Change looks fine, but what's up with the coverage/coveralls failure?

mbland:
Heh, one of those "Coveralls is comparing `floats` with `==` again" errors. Notice the `-0.0%` decrease, and clicking through reveals: "COVERAGE REMAINED THE SAME AT 78.106%". Coverage actually remained the same. Mind merging?

amoose:
To adjust that Coveralls threshold: 

> It looks like Coveralls allows a decrease threshold to be stated:

> screen shot 2016-01-30 at 8 13 44 pm

ccostino:
Ah nice, thanks for the tip @amoose!

Merging, @mbland.

mbland:
@amoose Where is that setting? I can't seem to find it.

PR #575, opened 2016-03-01
Copy edits to better communicate hub status

mtorres253:
changed _config.yml, index.html, header and footer to update messaging and links


mbland:
This should be re-opened against `18f-pages-internal` now.

PR #585, opened 2016-03-18
Redirects

nickbristow:
Ready after review


mtorres253:
Merged redirects! Let the games begin

